PR #17, opened 2018-08-03
terratest first pass

ryanking:
I have been playing with terratest a little bit to see if it would be useful for testing our modules.

In this diff I have 1 module with a bare-minimum test– basically a smoke test.

A couple things–

* test files are in the same directory as the module
* Since these tests are in go, you will now need to put your cztack checkout in your $GOPATH
* gonna try with terratest for now, will reevaluate once we get more experience

I will follow up with CI integration.


edulop91:
1. I like having the test along the code, easy to swap between them and tests usually have useful context. 
2. :+1: 
3. Agree with try it then decide

-------
1. :+1: 
2. :+1:

Gopkg.toml
```diff
@@ -0,0 +1,34 @@
+# Gopkg.toml example
```
edulop91: maybe rm these

Makefile
```diff
@@ -22,7 +22,11 @@ docs:
...
+		echo $$m; \
 		cd $$m; \
 		../scripts/update-readme.sh check || exit $$?; \
 		cd ..; \
-	done;
\ No newline at end of file
+	done;
+
+test: fmt
+	GOCACHE=off AWS_PROFILE=czi-si go test ./...
```
edulop91: why `GOCACHE=off`?

aws-iam-policy-cwlogs/module_test.go
```diff
@@ -0,0 +1,70 @@
...
+	"testing"
+
+	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+// An example of how to test the simple Terraform module in examples/terraform-basic-example using Terratest.
```
edulop91: old comment

Makefile
ryanking: go wil cache test results. cache is keyed off of go code. so if you change terraform but not go, it will not re-run the tests.

aws-iam-role-security-audit/module_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/gruntwork-io/terratest/modules/aws"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAWSIAMRoleReadOnly(t *testing.T) {
+	t.Parallel()
+
+	region := "us-west-1"
```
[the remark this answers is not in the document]
ryanking: Yeah I didn't do that because I didn't know where to put shared code, but since go will just treat this like a go project, we can create directory for shared test code.

PR #19, opened 2018-08-07
fix secrets reader policy modules + test

ryanking:


aws-params-secrets-setup/variables.tf
```diff
@@ -1 +1,5 @@
-
+variable "alias_name" {
+  description = "Chamber is hard coded to use a KMS alias with the name 'parameter_store_alias'."
```
edulop91: should that be `parameter_store_key` instead of `parameter_store_alias`

ryanking: Yeah

PR #23, opened 2018-08-09
move aurora modules

### LINKED ISSUE #24: move aurora modules

ryanking:
As part of this move, the modules have been refactored to have a common module. The -mysql and -postgres modules are still useful for setting appropriate defaults for each engine.

Fixes #24 #23

### DIFF
--- a/aws-aurora-mysql/outputs.tf
+++ b/aws-aurora-mysql/outputs.tf
@@ -1,7 +1,7 @@
 output "database_name" {
-  value = "${aws_rds_cluster.db.database_name}"
+  value = "${module.aurora.database_name}"
 }
 
 output "endpoint" {
-  value = "${aws_rds_cluster.db.endpoint}"
+  value = "${module.aurora.endpoint}"
 }
--- a/aws-aurora-mysql/outputs.tf
+++ b/aws-aurora-mysql/outputs.tf
@@ -1,0 +1,7 @@
+output "database_name" {
+  value = "${aws_rds_cluster.db.database_name}"
+}
+
+output "endpoint" {
+  value = "${aws_rds_cluster.db.endpoint}"
+}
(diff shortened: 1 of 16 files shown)


aws-aurora-postgres/main.tf
```diff
@@ -0,0 +1,87 @@
...
+}
+
+resource "aws_security_group" "rds" {
+  name        = "${local.name}"
+  description = "Allow postgres inbound"
+
+  vpc_id = "${var.vpc_id}"
+
+  ingress {
+    from_port   = "${var.database_port}"
```
edulop91: hmm maybe these shouldn't be vars. not currently propagated to the db resource.

edulop91:
I think we should refactor into a common module - there's already some code drift between them that we probably don't want

aws-aurora-postgres/main.tf
ryanking: That's weird, not sure how that would have worked before unless no one set them.

aws-aurora-mysql/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "database_name" {
+  value = "${module.aurora.database_name}"
+}
+
+output "endpoint" {
+  value = "${module.aurora.endpoint}"
+}
```
edulop91: maybe we should output the port too

aws-aurora-postgres/outputs.tf
```diff
@@ -0,0 +1,7 @@
+output "database_name" {
+  value = "${module.aurora.database_name}"
+}
+
+output "endpoint" {
+  value = "${module.aurora.endpoint}"
+}
```
edulop91: maybe we should output the port too

aws-aurora-mysql/outputs.tf
ryanking: sure

PR #27, opened 2018-08-09
add aws-redis-node

ryanking:


aws-redis-node/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+  default = "redis"
+}
+
+variable "port" {
+  type    = "string"
+  default = "6379"
+}
+
+variable "instance_type" {
+  default = "cache.m3.large"
```
edulop91: maybe make it `cache.m4.large`

PR #35, opened 2018-08-14
params refactor

ryanking:
A little bit different than planned, we now have–

* `aws-param` for reading a single param
* `aws-params-writer` for writing a batch of params for a service
* `aws-params-secrets-setup` for doing the per-account setup for chamber compat
* `aws-params-reader-policy` for creating policy to read secrets.


aws-params-writer/README.md
```diff
@@ -0,0 +1,34 @@
+Sets unencrypted string parameters in the AWS SSM parameter store. Often used in combination with Chamber
```
edulop91: should it be `encrypted`?

PR #42, opened 2018-08-21
Added GetParameter and GetParameterHistory to aws-params-reader-policy

chrisgoffinet:


aws-params-reader-policy/main.tf
```diff
@@ -20,7 +20,9 @@ resource "aws_iam_policy" "policy" {
     {
       "Sid": "",
       "Action": [
-        "ssm:GetParameters*"
+        "ssm:GetParameters*",
```
mbarrien: I personally hate wild cards (because new operations may share a wildcard name and give access to unexpected things) and would rather have GetParameters and GetParametersByPath explicit, but I won't let that block this PR.

PR #46, opened 2018-08-28
Output Readonly role ARN, fix bool

mbarrien:
This PR contains 2 separate small fixes:
1) Adds readonly role's ARN as output of the module. Getting the role ARNs is useful in some ECR IAM permission experiments.
2) Fixes the usage of boolean comparison in aws-param, to properly support both bool true and string "true".


edulop91:
update docs

PR #49, opened 2018-08-29
poweruser role:  Add more permissions to manage instanceprofile policies

edulop91:


edulop91:
Went back and forth with @travismccauley  a bit, these seem to be sufficient.

PR #52, opened 2018-08-30
init + apply test for aws-redis-node

### LINKED ISSUE #50: aws-redis-node apply test

ryanking:
This is our first test that uses one of our CI VPCs. The ids needed for this (vpc, subnets, etc) are being passed in as environment variables.

My thought is that if others want to fork and run the tests in their own infra, they can just set those env variables.

Fixes #50


testutil/util.go
```diff
@@ -78,3 +109,37 @@ func Cleanup(t *testing.T, options *terraform.Options) {
...
+	generator := newRand()
+	for i := 0; i < uniqueIDLength; i++ {
+		out.WriteByte(base62chars[generator.Intn(len(base62chars))])
+	}
+
+	return out.String()
+}
+
+// newRand creates a new random number generator, seeding it with the current system time.
+func newRand() *rand.Rand {
```
edulop91: idk if we'd ever want to print the seed to repro

ryanking: Yeah I can add that to the improvements we'd want to make to terratest, ok?

edulop91: :+1:

PR #55, opened 2018-09-04
Aurora module tests

### LINKED ISSUE #53: aws-aurora-postgres apply test

ryanking:
I set out on this change to write some full `apply` tests for the aurora modules. Along the way I had to made a bunch of small fixes to get things green.

Note that these are the first tests that use our CI VPC. I've decided, for  now, to just bake the relevant IDs into the Makefile in such a way that others an override them.

In theory the aurora tests should run in paralle (`t.Parallel`) but in practice that does not appear to be working and I don't know why.

Fixes #53 and #54

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ MODULES=$(filter-out vendor/ module-template/ scripts/ testutil/,$(sort $(dir $(
 TEST :=./...
 export PRIVATE_SUBNETS :=subnet-0e74698925a68c650,subnet-0f6ea862112b067c8
 export VPC_ID :=vpc-0442f170b88f8eaf6
+export VPC_CIDR_BLOCK :=10.72.0.0/16
+export DATABASE_SUBNET_GROUP :=shared-infra-cztack-ci
 
 all: clean fmt docs lint test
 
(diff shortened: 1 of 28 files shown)


Makefile
```diff
@@ -38,4 +40,4 @@ clean:
 		rm **/*.tfstate*
 
 test: fmt
-	GOCACHE=off AWS_PROFILE=cztack-ci-1 go test $(TEST)
+	GOCACHE=off AWS_PROFILE=cztack-ci-1 go test -parallel 10 -test.timeout 30m $(TEST)
```
edulop91:
```
Build cache requirement
Go 1.11 will be the last release to support setting the environment variable GOCACHE=off to disable the build cache, introduced in Go 1.10. Starting in Go 1.12, the build cache will be required, as a step toward eliminating $GOPATH/pkg. The module and package loading support described above already require that the build cache be enabled. If you have disabled the build cache to avoid problems you encountered, please file an issue to let us know about them.
```

edulop91: for sometime in the future

ryanking: ok thanks

PR #57, opened 2018-09-07
[feature] aws single page app static site + acm cert tests

edulop91:


aws-single-page-static-site/README.md
```diff
@@ -0,0 +1,17 @@
+# Module template
+
+## Example
+
+```hcl
+module "group" {
+  source = "github.com/chanzuckerberg/cztack/...?ref=v0.14.0"
```
ryanking: You should fill out the source and put in the var names

aws-single-page-static-site/README.md
```diff
@@ -0,0 +1,17 @@
+# Module template
```
ryanking: We should explain a bit about what this module is useful for? Like why is this specific to single page sites?

aws-single-page-static-site/module_test.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+func TestAwsSinglePageStaticSiteInit(t *testing.T) {
+	options := &terraform.Options{
+		TerraformDir: ".",
+	}
+	terraform.Init(t, options)
+}
+
+func TestAwsSinglePageStaticSiteInitAndApply(t *testing.T) {
```
edulop91: test is quite slow

aws-single-page-static-site/module_test.go
```diff
@@ -0,0 +1,54 @@
...
+	t.Skip("Skipping because destroy is painfully slow (>30m on average) - consider running destroy out of band")
```
edulop91: Skipping for now so we don't leak resources

PR #67, opened 2018-10-10
[breaking] Simplify Bless module

edulop91:


bless-ca/kms.tf
```diff
@@ -1,3 +1,5 @@
+data "aws_caller_identity" "current" {}
```
edulop91: i recently learned about this

PR #77, opened 2018-11-28
aws-iam-instance-profile Add option to disable attaching SSM policy

mbarrien:
This PR adds the ability to disable attaching the SSM policy to the role created in aws-iam-instance-profile. It is believed this will disable the ssm-agent, and specifically prevent it from creating the ssm-user on the local machine, which has been messing up some pre-set UIDs.

### DIFF
--- a/aws-iam-instance-profile/variables.tf
+++ b/aws-iam-instance-profile/variables.tf
@@ -14,3 +14,9 @@ variable "role_description" {
   description = "The description of the IAM role."
   default     = ""
 }
+
+variable "enable_ssm" {
+  type        = "string"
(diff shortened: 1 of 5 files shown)


aws-iam-instance-profile/variables.tf
```diff
@@ -14,3 +14,9 @@ variable "role_description" {
   description = "The description of the IAM role."
   default     = ""
 }
+
+variable "enable_ssm" {
```
edulop91: or disable_ssm?

edulop91: nvm

PR #79, opened 2018-12-04
Exposing the reader endpoint from the aws_rds_cluster resource.

huzefa:


mbarrien:
Please merge from master, then run make docs and check the changed files in. You may need to brew upgrade terraform-docs for this to work.

PR #90, opened 2019-04-22
Update aws-params-writer to explicitly take a parameters count

edulop91:


aws-params-writer/variables.tf
```diff
@@ -22,3 +22,8 @@ variable "parameters" {
   type        = "map"
   description = "Map from parameter names to values to set."
 }
+
+variable "parameters_count" {
+  type        = "string"
+  description = "HACK: The number of keys in var.parameters. To avoid hitting value of count cannot be computed."
```
SalmaCodes: Can you add a comment/TODO to remove this once we upgrade to Terraform 0.12?

It'd be nice to start keeping track of all the places our code will improve/change once we upgrade.

PR #92, opened 2019-04-24
[breaking] Removing ability of powerusers to register domains via route53

oliviabholmes:
Deny power users ability to register domains in Route53. 

Next step: create new user to specifically deal with domain registration/transfers in Route53

### DIFF
--- a/aws-iam-role-poweruser/main.tf
+++ b/aws-iam-role-poweruser/main.tf
@@ -85,6 +85,7 @@ data "aws_iam_policy_document" "misc" {
       "iam:DeleteUser",
       "iam:CreateGroup",
       "iam:DeleteGroup",
+      "route53domains:RegisterDomain",
     ]
 
     resources = ["*"]


austinylin:
🎉

aws-iam-role-poweruser/main.tf
```diff
@@ -85,6 +85,7 @@ data "aws_iam_policy_document" "misc" {
       "iam:DeleteUser",
       "iam:CreateGroup",
       "iam:DeleteGroup",
+      "route53domains:RegisterDomain",
```
edulop91: Would more generally denying `route53domains:*` be better?

edulop91: I guess it doesn't really matter if we don't have domains to start with.

oliviabholmes: I thought about that - but I think that the intention here was just to remove the ability of power users to register domains (because we want to force them to use MarkMonitor for that) and there's no real harm that I see in keeping the other route53domains permissions there. @austinylin do you have any thoughts?

oliviabholmes:
@edulop91 Not sure why this would cause an error in the build... 

The error is `The command "make check-docs" exited with 2.`

ryanking:
@oliviabholmes it seems that the build has been failing for awhile, independent of this change.

ryanking:
once https://github.com/chanzuckerberg/cztack/pull/93 is in, merge master and your build should be green.

oliviabholmes:
@edulop91 Not sure what the team's policy is on merging your own PRs but the build is passing now!

ryanking:
@oliviabholmes add the tag "bot/merge" and @czimergebot will merge it for you

PR #96, opened 2019-05-15
Adding in Route53ReadOnly Access to Route53Domains-PowerUser Role

oliviabholmes:
Needed to add the route53ReadOnly policy to the Route53 power user IAM role that I created previously. I need to be able to read and download the zone files for our hosted domains and this should give me the privileges to do that.


oliviabholmes:
cc: @austinylin

PR #101, opened 2019-06-13
WIP: Create region-bound poweruser role

edulop91:
### Summary
Opening this PR draft to get some early feedback before I invest more time going down this path.
The idea is to limit the surface area powerusers can interact with by:
- constraining regions available
- constraining aws services available

risks:
- [iam limits](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html)

### Test Plan
- test in some of our roles

### References
- https://aws.amazon.com/blogs/security/easier-way-to-control-access-to-aws-regions-using-iam-policies/

### DIFF
--- a/aws-iam-role-poweruser/lambda.tf
+++ b/aws-iam-role-poweruser/lambda.tf
@@ -1,0 +1,57 @@
+
+locals {
+  enable_lambda = "${var.enable_lambda ? 1 : 0}"
+
+}
+// Readonly everywhere
+resource "aws_iam_role_policy_attachment" "lambda-readonly" {
+  role       = "${aws_iam_role.poweruser.name}"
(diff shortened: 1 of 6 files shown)


aws-iam-role-poweruser/lambda.tf
```diff
@@ -0,0 +1,57 @@
...
+
+}
+// Readonly everywhere
+resource "aws_iam_role_policy_attachment" "lambda-readonly" {
+  role       = "${aws_iam_role.poweruser.name}"
+  policy_arn = "arn:aws:iam::aws:policy/AWSLambdaReadOnlyAccess"
+}
+
+
+// Write confined to regions
```
ryanking: Rather than duplicating the AWS-managed policy, why don't we combine two policies

1. the aws-managed one
2. one of ours that does a deny on StringNotEquals region

?

PR #104, opened 2019-07-22
[breaking] Changes for TF0.12 compatibility without break TF0.11 usage

mbarrien:
### Summary
This PR contains changes that make cztack closer to 0.12 compatible while remaining compatible with 0.11. They mostly are mutually-compatible syntactic changes although may require code changes because of variable name changes.

These include:
* Using go modules instead of the standalone dep tool
* Ensuring that map variables have an equal sign
* Renaming the depends_on variable in some modules to dependencies (corresponding code change required in repos using those modules)
* Compatible syntax for relative directories (0.12 requires "./" to be at the start, even when using a parent directory)


aws-iam-group-assume-role/main.tf
```diff
@@ -32,3 +34,22 @@ resource "aws_iam_group_policy_attachment" "assume-role" {
   policy_arn = "${aws_iam_policy.assume-role.arn}"
   group      = "${aws_iam_group.assume-role.name}"
 }
+
+# # https://github.com/hashicorp/terraform/issues/1178
+
+# # Pseudo dependencies to ensure dependencies actually get enforced
```
edulop91: why are these commented out?

mbarrien: Although this workaround as described in the comments on the above issue is actually semantically correct, in practice the need to shell out makes Terraform complain about running out of file handles. I wanted something in here that shows that a) We currently do this incorrectly which is why we have to double apply whenever we add a new user, and b) here's semantically what we want it to do.

aws-aurora/main.tf
```diff
@@ -21,7 +21,7 @@ resource "aws_security_group" "rds" {
     from_port   = "${var.port}"
     to_port     = "${var.port}"
     protocol    = "tcp"
-    cidr_blocks = ["${var.ingress_cidr_blocks}"]
```
edulop91: these do break my tf 0.11. just ran into a couple instances

aws-iam-group-assume-role/main.tf
edulop91: maybe add a comment to that effect in the code? otherwise having that code commented out there doesn't give us much

aws-aurora/main.tf
mbarrien: Moved the whole commit to the 0.12 breaking PR.

aws-iam-group-assume-role/main.tf
mbarrien: It's in the comment of the commit, not in the code itself. Good enough?

mbarrien: Actually added the comment to the code, since I realize the ensuing squash will remove it.

PR #105, opened 2019-07-22
[breaking] Terraform 0.12 compatibility (breaks 0.11 compatibility)

mbarrien:
### Summary
This PR adds the changes over and above #104 to make cztack compatible with 0.12. Specifically:
* Removing brackets around arguments that are already supposed to be lists. In some rare cases (involving calculated fields), this doesn't work properly in 0.11, so considering it breaking.
* Removes use of terraform-docs; terraform-docs is incompatible with Terraform 0.12 (see https://github.com/segmentio/terraform-docs/issues/62)
* Uses the new dynamic block syntax to be able to pass dynamically generated blocks that coincidentally worked in 0.11 but are incompatible with 0.12.

### Test Plan
Testing via corresponding shared-infra PR successfully passing.


edulop91:
Worth specifying in the README tf >= 0.12

mbarrien:
Added README comment about TF0.12 needed.

PR #111, opened 2019-08-10
Add aws-ssm-params and aws-ssm-params-writer

mbarrien:
### Summary
This PR adds a Terraform 0.12 compatible aws-ssm-params and aws-ssm-params-writer module. It is intended to replace the aws-params and aws-params-writer module. We did not just modify the existing one because the migration from the old Terraform state to the new Terraform state is non obvious and its usage is widespread and thus likely to cause pain. The plan is to deprecate aws-params and aws-params-writer.

The big difference between the old and new version of aws-ssm-params-writer is the use of the for_each resource construct in Terraform 0.12. This removes the need for the parameters_count input, at the cost of loss of backwards compatibility with the existing module.

aws-ssm-params is now capable of reading multiple SSM parameters and returning them all as a single map output. The list of variables must be specified as an input; there's no simple way yet to just read all variables under a given path inside Terraform.


aws-ssm-params/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```hcl
+module "secret" {
+  source = "github.com/chanzuckerberg/cztack/aws-ssm-params-secret?ref=v0.14.0"
+
+  project = "acme"
+  env     = "staging"
+  service = "website"
+
+  name = "password"
```
redhoteggplant:
```suggestion
  parameters = "password"
```
and `module.secret.values` in the output?

mbarrien: Fixed.

PR #116, opened 2019-09-10
Add Aurora deletion protection variable

chauvm:


aws-aurora/main.tf
```diff
@@ -50,6 +50,7 @@ resource "aws_rds_cluster" "db" {
   kms_key_id                          = "${var.kms_key_id}"
   port                                = "${var.port}"
   db_cluster_parameter_group_name     = "${aws_rds_cluster_parameter_group.db.id}"
+  deletion_protection                 = "${var.db_deletion_protection}"
```
edulop91: defaulting this to true might break some of the tests in this repo, the tests might need to specify a false

chauvm: hmm I'll change the default to false then, since I need this one now. But I think deletion protection should be true by default.

chauvm: I'll change it if the tests fail

edulop91: yeah, i think deletion protection can stay true, but the terraform in the tests need to specify false

edulop91: like https://github.com/chanzuckerberg/cztack/blob/master/aws-aurora-postgres/module_test.go#L51

edulop91: you might also need to go manually delete these databases since the test will probably not be able to delete them

PR #132, opened 2019-09-23
[feature] Add ECS modules

mbarrien:
### Summary
Adds ECS service (public facing with a frontend) and ECS job (internal headless applications) modules to the public cztack repo, with separate modules for Fargate and non-Fargate jobs.

Changes relative to internal repo:
* Uses `aws_lb_*` resources instead of `aws_alb_*` resources
* Separate modules for fargate and EC2 jobs. EC2 tasks default to host networking and the host's security group, but we add support for awsvpc networking that assigns an ENI with its own security group to the task (Fargate always has this.)
* Variable names do not have `fargate_*` prefix since module is Fargate-specific.
* We now have a variable called `manage_task_definitions` instead of relying on a non-empty/non-null task_definition variable. This gets rid of the problem with "count cannot be determined from compute value" errors, but now requires the user to properly specify manage_task_definitions for their use case.
* Supports source security groups for the ingress LB. This will allow us to limit services by security group instead of just CIDR blocks, which is useful when you have a service that only should ever talk to a single upstream service (e.g. a proxy)
* Adds AWS tags to the ECS service and task definitions.


ryanking:
> * Separate modules for fargate and EC2 jobs (note that although theoretically possible, ECS EC2 module does not support

?

aws-ecs-job-fargate/README.md
```diff
@@ -0,0 +1,45 @@
...
+This creates an ECS service running on Fargate with no load balancer in front of it. Good for
+background worker daemon sort of things.
+
+## Terraform managed task definition vs czecs
+
+If the user sets var.manage_task_definition = true, Terraform will manage the lifecycle
+of the container definition; any external changes are reset on the next Terraform run.
+
+If var.manage_task_definition = false, the user is expected to manage the
+container definition external to Terraform (e.g. using czecs). Upon creation,
```
ryanking:
```suggestion
container definition external to Terraform (e.g. using [czecs](https://github.com/chanzuckerberg/czecs)). Upon creation,
```

mbarrien:
Fixed PR description

aws-ecs-service/README.md
```diff
@@ -0,0 +1,183 @@
+# ECS Service
+
+This module creates an ECS service fronted by a public ALB.
+
+## Terraform managed task definition vs czecs
+
+If the user sets var.manage_task_definition = true, Terraform will manage the lifecycle
```
ryanking:
```suggestion
If the user sets `var.manage_task_definition = true`, Terraform will manage the lifecycle
```

aws-ecs-job-fargate/main.tf
```diff
@@ -0,0 +1,144 @@
...
+  deployment_maximum_percent         = var.deployment_maximum_percent
+  deployment_minimum_healthy_percent = var.deployment_minimum_healthy_percent
+  scheduling_strategy                = "REPLICA"
+
+  network_configuration {
+    subnets         = var.task_subnets
+    security_groups = var.security_group_ids
+  }
+
+  lifecycle {
```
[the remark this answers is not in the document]
mbarrien: 2 different attempts show it's not.
With...
```
  lifecycle {
    ignore_changes = compact([var.manage_task_definition ? "" : task_definition])      
  }
```
... I get:
```
Error: Invalid expression

  on .terraform/modules/golinks.oauth-service.proxy.service/aws-ecs-service-fargate/service.tf line 70, in resource "aws_ecs_service" "job":
  70:     ignore_changes = compact([var.manage_task_definition ? "" : task_definition])

A static list expression is required.
```
(It underlined "compact" so it's not about task_definition in the list)
---

With...
```
  dynamic "lifecycle" {
    for_each = compact([var.manage_task_definition ? "" : "present"])
    content {
      ignore_changes = [task_definition]
    }
  }
````
...I get  the following:
```
Error: Unsupported block type

  on .terraform/modules/golinks.oauth-service.web-service/aws-ecs-service-fargate/service.tf line 69, in resource "aws_ecs_service" "job":
  69:   dynamic "lifecycle" {

Blocks of type "lifecycle" are not expected here.
```

mbarrien:
Added untested support for awsvpc networking for non-Fargate jobs, via var.awsvpc_network_mode flag. Also added secrets manager for Docker image fetching. Note that using that flag also means that jobs using czecs would need to add networkMode and/or execution role to the definition.

PR #139, opened 2019-10-04
[feature] aws-ecs Add support for ordered_placement_strategy

mbarrien:


mbarrien:
We were able to avoid needing to actually need this feature in our CZI code base, but it may still be good to have this feature available, in case we want to use ECS to enforce spreading ECS tasks onto different machines. Please review.

aws-ecs-job-fargate/variables.tf
```diff
@@ -88,4 +88,10 @@ variable "manage_task_definition" {
   description = "If false, Terraform will not touch the task definition for the ECS service after initial creation"
   type        = bool
   default     = true
+}
+
+variable "ordered_placement_strategy" {
+  type        = list(object({ type = string, field = string }))
```
edulop91: neat

PR #140, opened 2019-10-06
[fix] aws-ecs support not applying tags to services

mbarrien:
### Summary
AWS added support for tagging ECS services in a way that is not backwards compatible with ECS services that have been deployed for a long time. Tags can only be added to ECS services that use a new ARN format that embeds a cluster name inside of the ARN, but older ECS service do not have the name embedded.

For some teams, there are many services that are already deployed, and replacing the ECS service just to apply tags will result in user-visible downtime. We provide a flag for users migrating from the CZI internal module to this public cztack module to be able to prevent Terraform from trying to apply tags, in those cases where the user has determined that applying tags is incompatible with the existing resource. It is recommended that at the next possible down time, that the service be replaced by running terraform taint, then restoring the ECS task definition version (since the taint/replace will restore to only the last stub definition). After the service is destroyed/replaced, the tag_service=false flag can be removed.


ryanking:
@mbarrien can we add the note in the PR description to the README for this module?

mbarrien:
Added more refined wording to all the readme files. Please re-review.

PR #163, opened 2019-12-31
Support EFS

jgadling:
### Summary
Support creation of EFS filesystems in cztack

### Test Plan
Once this is merged, I'll add an EFS drive to playground and try to get it mounted to some EC2 instances.

### DIFF
--- a/aws-efs-volume/module_test.go
+++ b/aws-efs-volume/module_test.go
@@ -1,0 +1,35 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/chanzuckerberg/cztack/testutil"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+)
+
+func TestPrivateBucket(t *testing.T) {
+
+	project := testutil.UniqueId()
+	env := testutil.UniqueId()
+	service := testutil.UniqueId()
+	owner := testutil.UniqueId()
+	vpcId := testutil.UniqueId()
+
+	volumeName := testutil.UniqueId()
+
+	options := testutil.Options(
+		testutil.DefaultRegion,
+		map[string]interface{}{
+			"project": project,
+			"env":     env,
+			"service": service,
+			"owner":   owner,
+
+			"volume_name": volumeName,
+			"vpc_id": vpcId,
+		},
+	)
+
+	defer terraform.Destroy(t, options)
+	testutil.Run(t, options)
+}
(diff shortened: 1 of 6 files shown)


ryanking:
@jgadling  you should be able to test this before it is merge as well

aws-efs-volume/module_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/gruntwork-io/terratest/modules/terraform"
+)
+
+func TestPrivateBucket(t *testing.T) {
+
+	project := testutil.UniqueId()
+	env := testutil.UniqueId()
+	service := testutil.UniqueId()
+	owner := testutil.UniqueId()
+	vpcId := testutil.UniqueId()
```
ryanking: You are going to need to use an id for a real VPC here. There is one set as a variable in the Makefile https://github.com/chanzuckerberg/cztack/blob/master/Makefile#L5

aws-efs-volume/outputs.tf
```diff
@@ -0,0 +1,12 @@
+// HACK(el): we do this to hint TF dependency graph since modules can't depend_on
+output "id" {
+  value = aws_efs_file_system.efs.id
+}
+
+output "dns_name" {
```
ryanking: Will there be different dns names per subnet?

aws-efs-volume/module_test.go
```diff
@@ -0,0 +1,35 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/chanzuckerberg/cztack/testutil"
+	"github.com/gruntwork-io/terratest/modules/terraform"
+)
+
+func TestPrivateBucket(t *testing.T) {
```
chauvm: maybe rename this function

aws-efs-volume/outputs.tf
jgadling: According to some blog posts, the magic VPC DNS resolution should make the same name work (and resolve to different IP's) in all subnets, but I'm going to validate that. Seems fishy to me.

jgadling: Learned more when testing this module further: Even though the mount point in each subnet has its own IP, there's only one DNS name for the volume. So we only need to output one DNS name from the module.

The individual mount point IP's *could* be useful outputs if someone's getting tricky with networking, but imnsho we can keep the interface simple for now and expand it if we need more detail.

jgadling:
Pretty sure this thing is failing due to some KMS replication delay unrelated to my new module: 
https://travis-ci.com/chanzuckerberg/cztack/jobs/271482796

PR #168, opened 2020-01-13
[breaking] [feature] allow setting ca_cert_identifier on aurora modules

ryanking:
This will allow upgrading aurora database instances to the new certificate authority.

* https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html

### DIFF
--- a/aws-aurora-mysql/terraform.tf
+++ b/aws-aurora-mysql/terraform.tf
@@ -1,0 +1,6 @@
+terraform {
+  required_providers {
+    # ca_cert_identifier on RDS was added in 2.44.0
+    aws = ">= 2.44.0"
+  }
+}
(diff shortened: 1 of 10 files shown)


aws-aurora-mysql/terraform.tf
```diff
@@ -0,0 +1,6 @@
+terraform {
```
edulop91: neat, didn't know about this

aws-aurora-mysql/variables.tf
```diff
@@ -167,3 +167,9 @@ variable "engine_version" {
   type    = string
   default = "5.7"
 }
+
+variable ca_cert_identifier {
+  type        = string
+  description = "Identifier for the certificate authority. Use rds-ca-2015 for anything new."
```
edulop91:
```suggestion
  description = "Identifier for the certificate authority. rds-ca-2015 is the latest available version."
```
not sure how I feel about that change, but maybe?

aws-aurora-mysql/terraform.tf
ryanking: me either!

PR #173, opened 2020-02-07
allow aws-s3-private-bucket take in custom lifecycle_rule

michaelmcchen:
### Summary
This change allows an implementation of `aws-s3-private-bucket` to use the current lifecycle_rule, or provide a different one via a TF variable. For example, in my implementation I'd like to have a shorter 30-day TTL on "current" objects in the bucket. 

This is a little cumbersome, but I'm unsure if there's a cleaner way (new to TF lol) of doing this, since the `s3` resource doesn't take in a named argument but rather a configuration block. 

### Test Plan
Verified via `make plan` on a resource. 

### References

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -20,10 +20,10 @@ resource "aws_s3_bucket" "bucket" {
 
   # dynamic block used instead of simply assigning a variable b/c lifecycle_rule is configuration block
   dynamic "lifecycle_rule" {
-    for_each = var.lifecycle_rule
+    for_each = var.lifecycle_rules
 
     content {
-      id                                     = lookup(lifecycle_rule.value, "id", null) #lookup() provides default value in case it does not exist in var.lifecycle_rule input
+      id                                     = lookup(lifecycle_rule.value, "id", null) #lookup() provides default value in case it does not exist in var.lifecycle_rules input
       prefix                                 = lookup(lifecycle_rule.value, "prefix", null)
       tags                                   = lookup(lifecycle_rule.value, "tags", null)
       enabled                                = lookup(lifecycle_rule.value, "enabled", false)
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -18,11 +18,12 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.enable_versioning
   }
 
+  # dynamic block used instead of simply assigning a variable b/c lifecycle_rule is configuration block
   dynamic "lifecycle_rule" {
     for_each = var.lifecycle_rule
 
     content {
-      id                                     = lookup(lifecycle_rule.value, "id", null)
+      id                                     = lookup(lifecycle_rule.value, "id", null) #lookup() provides default value in case it does not exist in var.lifecycle_rule input
       prefix                                 = lookup(lifecycle_rule.value, "prefix", null)
       tags                                   = lookup(lifecycle_rule.value, "tags", null)
       enabled                                = lookup(lifecycle_rule.value, "enabled", false)
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -18,26 +18,54 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.enable_versioning
   }
 
-  lifecycle_rule {
-    enabled = true
+  dynamic "lifecycle_rule" {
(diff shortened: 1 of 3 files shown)


aws-s3-private-bucket/main.tf
```diff
@@ -18,26 +18,54 @@ resource "aws_s3_bucket" "bucket" {
     enabled = var.enable_versioning
   }
 
-  lifecycle_rule {
-    enabled = true
+  dynamic "lifecycle_rule" {
+    for_each = var.lifecycle_rule
+
+    content {
+      id                                     = lookup(lifecycle_rule.value, "id", null)
```
ryanking: I think you can "just" do `lifecycle_rule.value.id`, right? If that's not the case then we should leave a comment about why we need to do this.

aws-s3-private-bucket/variables.tf
```diff
@@ -34,3 +34,26 @@ variable "abort_incomplete_multipart_upload_days" {
   description = "Number of days after which an incomplete multipart upload is canceled."
   default     = 14
 }
+
+variable "lifecycle_rule" {
+  description = "List of maps containing configuration of object lifecycle management."
+  type        = any
```
ryanking: Interesting– it had not occurred to me that we can't use structural typing here, since we want to allow values to be be absent - https://www.terraform.io/docs/configuration/types.html#structural-types

aws-s3-private-bucket/main.tf
michaelmcchen: That works, but not all `lifecycle_rule` properties are present in the default rule (in `variables.tf`). Thus if I did `lifecycle_rule.value.prefix`, Terraform would complain that it couldn't find it, hence the `lookup()` w/ the default values. 

I can definitely do the short-hand for the properties that are present in the default rule and keep the `lookup()` for those that aren't present, but it feels very un-uniform. Also, it's possible the default rule might change in the future - we wouldn't want to couple the two together. Thoughts?

ryanking: Yeah I see that now (sorry this is a new feature and I don't know all the subtleties yet).

Given that constraint I think it is fine to leave as-is until we find a more elegant way to do it.

ryanking:
> @ryanking Also, do I need to make changes to `README.md`? I tried doing `make docs` but I got the following:

You need to install [terraform-docs](https://github.com/segmentio/terraform-docs).

aws-s3-private-bucket/main.tf
palasha: > for those that aren't present, but it feels very un-uniform.

Personally, I think that would be better. It signals intent clearly to the reader whether the programmer expects this value to exist, or that it might not.

> we wouldn't want to couple the two together. Thoughts?

Then we would also change its callsites to reflect the new signature?

aws-s3-private-bucket/variables.tf
palasha: After reading those docs, it's not clear to me why we couldn't do this? Could it be typed as `list(object(...))` and then have the default be an empty list?

michaelmcchen:
> for this specific use case which will be very common with our s3 buckets

@palasha I think I would rephrase this as "the idea of putting a TTL on objects" will be very common, but the lifecycle rule API provides a decent amount of configuration - ie. apply on a certain path prefix, expire by a certain date, etc. I'm unsure if we should hide that from users, since it's flexibility that could be very useful.

aws-s3-private-bucket/main.tf
michaelmcchen: > whether the programmer expects this value to exist, or that it might not.

For most of these values, their existence depends on the configuration that you're trying to use right? IMO the default configuration provided in `variables.tf` shouldn't influence how it's being written here - what if we have add multiple default lifecycle_rules? 

I think signal to the programmer can be achieved via running `make plan` right?

aws-s3-private-bucket/variables.tf
michaelmcchen: If we use `object(...)` we'd have to define the named keys, which might be absent right? If we have the default be an empty list, how would we pass in the existing rule?

aws-s3-private-bucket/variables.tf
```diff
@@ -34,3 +34,26 @@ variable "abort_incomplete_multipart_upload_days" {
   description = "Number of days after which an incomplete multipart upload is canceled."
   default     = 14
 }
+
+variable "lifecycle_rule" {
```
ryanking: I think this should be plural, yeah?

PR #181, opened 2020-03-24
Update default certs to rds-ca-2019, update docs

mbarrien:
### Summary
The rds-ca-2015 certificate expired on March 5, 2020, and terratest runs that rely on this certificate fail with a non-existent certificate; we change the default certificates here. Although technically this is breaking for anything relying on default values, since the cert is already expired those uses are already broken anyways. While here, we also fix the terraform-docs of several unrelated modules.

### Test Plan
Existing unit tests pass.

### References
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html


mbarrien:
Was on an older branch, already fixed in master.

PR #183, opened 2020-03-24
[breaking] [feature] aws-iam-instance-profile reuse existing role and remove SSM support

mbarrien:
### Summary
* Allows passing an existing role to aws-iam-instance-profile, and the module will attach whatever permissions are needed. This allows using roles that are created and managed by other modules (e.g. EKS worker role), but provides ability to attach a standard set of permissions to that role. It will still always create the instance profile, even if the profile will be unused.
* Removes unused permissions from the aws-iam-instance-profile module for using SSM.

### Test Plan
Already in use in CZI internal playground deployment.

### DIFF
--- a/aws-iam-instance-profile/README.md
+++ b/aws-iam-instance-profile/README.md
@@ -39,7 +39,6 @@ resource "aws_instance" "instance" {
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:-----:|
-| enable\_ssm | Attach the appropriate policies to allow the instance to integrate with AWS Systems Manager. | `string` | `"true"` | no |
 | iam\_path | The IAM path to the role. | `string` | `"/"` | no |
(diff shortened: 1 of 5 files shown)


aws-iam-instance-profile/README.md
```diff
@@ -39,7 +39,8 @@ resource "aws_instance" "instance" {
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:-----:|
-| enable\_ssm | Attach the appropriate policies to allow the instance to integrate with AWS Systems Manager. | `string` | `"true"` | no |
+| create\_role | Creates a role for use with the instance profile. | `bool` | `true` | no |
```
ryanking: Instead of having a a second parameter, can we use the fact that `exiting_role_name` is null as a signal to create a new role?

PR #185, opened 2020-03-31
[feature] aws-iam-role-*: Allow multiple accounts to assume role, deprecate source_account_id

mbarrien:
Adds a source_account_ids field to aws-iam-role-* modules, which will allow multiple accounts' root to assume role, assuming the user/role in the other accounts allow the role assumption.

Either source_account_id or source_account_ids or both work; the intent is to keep backwards compatibility for now, but potentially eliminate source_account_id. Officially document that source_account_id is deprecated.

### DIFF
--- a/aws-iam-role-poweruser/variables.tf
+++ b/aws-iam-role-poweruser/variables.tf
@@ -9,6 +9,12 @@ variable "source_account_id" {
   description = "The source AWS account to establish a trust relationship. Ignored if empty or not provided."
 }
 
+variable "source_account_ids" {
+  type        = set(string)
+  default     = []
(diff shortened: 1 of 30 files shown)


aws-iam-role-poweruser/variables.tf
```diff
@@ -9,6 +9,12 @@ variable "source_account_id" {
   description = "The source AWS account to establish a trust relationship. Ignored if empty or not provided."
 }
 
+variable "source_account_ids" {
```
edulop91: worth adding a deprecation notice on the source_account_id ?

edulop91: oh also, `make docs`

mbarrien: Done on both.

PR #186, opened 2020-04-03
Expose the cluster endpoint in the aurora module

huzefa:
### Summary
- Exposing the aurora cluster endpoint so that we can attach a different class of instance to the cluster.
- Exposing the `db_parameter_group_name` which is also required when externally creating a new instance.
- Fix a bug where the module was passing the `engine_version` was being passed as the `params_engine_version` as well. This works when they're both `5.7` or some other select values but not always. For example, if version is `5.7.28` or any other minor version, that would be considered an invalid `params_engine_version` value.

### Test Plan
Tested this code directly by pulling this module into my existing source.

### DIFF
--- a/aws-aurora-mysql/README.md
+++ b/aws-aurora-mysql/README.md
@@ -71,6 +71,7 @@ No provider.
 | Name | Description |
 |------|-------------|
 | database\_name | n/a |
+| db\_parameter\_group\_name | n/a |
 | endpoint | n/a |
 | rds\_cluster\_id | n/a |
 | reader\_endpoint | n/a |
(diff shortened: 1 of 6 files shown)


aws-aurora-mysql/README.md
```diff
@@ -57,6 +57,7 @@ No provider.
 | instance\_count | Number of instances to create in this cluster. | `string` | `1` | no |
 | kms\_key\_id | If provided, storage will be encrypted with this key, otherwise an AWS-managed key is used. (Encryption is always on). | `string` | `""` | no |
 | owner | Owner for tagging and naming. See [doc](../README.md#consistent-tagging). | `string` | n/a | yes |
+| params\_engine\_version | n/a | `string` | `"5.7"` | no |
```
jgadling, line 60: mind including a description for this parameter?

ryanking: Note that you have to write the description in the terraform code and run `make docs`.

PR #188, opened 2020-04-15
[feature] make public_access_block on private bucket optional

ryanking:
In theory, for every private s3 bucket we should turn on the public access blocks.

### Test Plan
* CI tests

### References

### DIFF
--- a/aws-s3-private-bucket/variables.tf
+++ b/aws-s3-private-bucket/variables.tf
@@ -57,3 +57,8 @@ variable "lifecycle_rules" {
     }
   ]
 }
+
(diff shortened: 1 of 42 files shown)


aws-s3-private-bucket/variables.tf
```diff
@@ -57,3 +57,8 @@ variable "lifecycle_rules" {
     }
   ]
 }
+
+variable public_access_block {
```
edulop91, line 61: make docs

ryanking: derp

PR #194, opened 2020-05-20
[feature] Optionally disable poweruser role iam actions

edulop91:
### Summary
Occasionally we might want to disable the extra iam actions in poweruser role, make it configurable.

### DIFF
--- a/aws-iam-role-poweruser/README.md
+++ b/aws-iam-role-poweruser/README.md
@@ -31,6 +31,7 @@ No requirements.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
+| authorize\_iam | Indicates if we should augment the PoserUserAccess policy with certain IAM actions. | `bool` | `true` | no |
 | iam\_path | n/a | `string` | `"/"` | no |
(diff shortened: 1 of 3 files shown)


aws-iam-role-poweruser/variables.tf
```diff
@@ -38,3 +38,9 @@ variable oidc {
   default     = []
   description = "A list of AWS OIDC IDPs to establish a trust relationship for this role."
 }
+
+variable authorize_iam {
+  type        = bool
+  default     = true
+  description = "Indicates if we should augment the PoserUserAccess policy with certain IAM actions."
```
edulop91, line 45:
```suggestion
  description = "Indicates if we should augment the PowerUserAccess policy with certain IAM actions."
```

aws-iam-role-poweruser/README.md
```diff
@@ -31,6 +31,7 @@ No requirements.
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
+| authorize\_iam | Indicates if we should augment the PoserUserAccess policy with certain IAM actions. | `bool` | `true` | no |
```
edulop91, line 34:
```suggestion
| authorize\_iam | Indicates if we should augment the PowerUserAccess policy with certain IAM actions. | `bool` | `true` | no |
```

PR #199, opened 2020-05-27
[feature] standard tags for IAM roles

mbarrien:
Adds ability to add standard AWS tags to IAM roles. Tag arguments may be omitted for backwards compatibility.


mbarrien:
Substantial re-write to take a generic "tags" field instead of project/env/service. Change aws-iam-role-crossacct to take a variable by the name of "tags" instead of "role_tags" (didn't seem to be used in cztack or CZI's code base). Please re-review.

PR #201, opened 2020-06-08
[fix] move build to github actions

ryanking:
travis-ci has proven to be too unreliable and difficult to maintain. It seems like GHA will be a better thing for us to build on.

I toyed with a few approaches, but decided to run each module as a separate job. The setup time is pretty fast and having a separate status check for each module is pretty nice.

### Test Plan
* CI

### References
*


.github/workflows/ci.yml
```diff
@@ -0,0 +1,100 @@
...
+  #   steps:
+  #   - uses: actions/setup-go@v2
+  #     with:
+  #       go-version: 1.14.3
+  #   - run: make check-docs
+  test:
+    strategy:
+      matrix:
+        module: [
+          aws-acm-cert,
```
edulop91, line 36: these have to be statically defined?

PR #205, opened 2020-06-23
[feature] add grant dynamic block to s3 bucket to enable the option for using g…

Dingying0410:
### Motivation  
- We have referenced this bucket in https://github.com/FB-PLP/terraform-infra-management/blob/master/terraform/envs/staging/cloudfront-dev-summitlearning/cloudfront.tf as a logging bucket for Cloudfront distribution. 

- Recently I found that  when I run `make plan` in this  directory: it shows the following for logging bucket.
```
  # module.cloudfront-logging-bucket.aws_s3_bucket.bucket will be updated in-place
  ~ resource "aws_s3_bucket" "bucket" {
      .......
      - grant {
          - id          = "LP's_aws_canonical_user_id" -> null
          - permissions = [
              - "FULL_CONTROL",
            ] -> null
          - type        = "CanonicalUser" -> null
        }
      - grant {
          - id          = "f607bc3a24521bc2de593c73569197969569a247af9a1b615fd9d4a5ec6ed28f" -> null
          - permissions = [
              - "FULL_CONTROL",
            ] -> null
          - type        = "CanonicalUser" -> null
        }

     ......

        server_side_encryption_configuration {
            rule {
                apply_server_side_encryption_by_default {
                    sse_algorithm = "AES256"
                }
            }
        }

        versioning {
            enabled    = true
            mfa_delete = false
        }
    }
```

Then I figured out that  t is because to write logs to the bucket, we need to provide the permissions for Cloudfront  to write to the bucket
per [this article](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html) 
> When you create or update a distribution and enable logging, CloudFront uses these permissions to update the ACL for the bucket to give the awslogsdelivery account FULL_CONTROL permission. The awslogsdelivery account writes log files to the bucket. 

Here the `f607bc3a24521bc2de593c73569197969569a247af9a1b615fd9d4a5ec6ed28f` is the canonical id of the `awslogsdelivery`, and it gives the `awslogsdelivery` the permissions to write to this bucket. 

This motivated me to figure out how to make the `make plan` of  the logging bucket up to date. 

### Approach
Create a dynamic block for `grant` in this module.

I have been  thinking about other ways like
- using a `aws_s3_bucket` instead of referencing the module in the https://github.com/FB-PLP/terraform-infra-management/blob/master/terraform/envs/staging/cloudfront-dev-summitlearning/cloudfront.tf, with several `state mv` 
but this would make us discard the other benefits provided by this module 

### Test plan: 
1. add the following changes in `cloudfront-dev-summitlearning/cloudfront.tf`: 
```
module "cloudfront-logging-bucket" {
  source      = "git@github.com:chanzuckerberg/cztack.git//aws-s3-private-bucket?ref=christina/add_grant_to_s3_bucket"
  bucket_name = "${var.project}-${var.env}-${var.component}-logs"
  env         = var.env
  owner       = var.owner
  project     = var.project
  service     = var.component

  grants = [
    {
      canonical_user_id : "LP's_aws_canonical_user_id" # 
      permissions : ["FULL_CONTROL"]

    },
    {
      canonical_user_id : "f607bc3a24521bc2de593c73569197969569a247af9a1b615fd9d4a5ec6ed28f" # Log delivery group canonical id 
      permissions : ["FULL_CONTROL"]

    }
  ]
}
```
2. after I added this code, `make plan` is up to date, the differences in the PR summary disappeared  

### Next plan
After this PR gets applied, update the code in all the cloudfront logging  buckets that LP has set up:
- https://github.com/FB-PLP/terraform-infra-management/blob/master/terraform/envs/staging/cloudfront-dev-summitlearning/cloudfront.tf
- https://github.com/FB-PLP/terraform-infra-management/blob/master/terraform/envs/staging/cloudfront-latest-summitlearning/cloudfront.tf
- https://github.com/FB-PLP/terraform-infra-management/blob/master/terraform/envs/produciton/cloudfront-prod-summitlearning/cloudfront.tf

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -1,5 +1,5 @@
 locals {
-  use_grant_acl = length(var.canonical_user_id_and_permissions) == 0 ? false : true
+  acl = length(var.canonical_user_id_and_permissions) == 0 ? null : "private"
 
   tags = {
     project   = var.project
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -1,4 +1,6 @@
 locals {
+  use_grant_acl = length(var.canonical_user_id_and_permissions) == 0 ? false : true
+
   tags = {
     project   = var.project
     env       = var.env
@@ -10,7 +12,31 @@ locals {
 
 resource "aws_s3_bucket" "bucket" {
   bucket = var.bucket_name
-  acl    = "private"
+  # To define the access control list (ACL) of the bucket, there are two ways:
+  # 1. Define the canned ACL using the acl argument
+  # 2. Define the grant ACL using the grant argument
+  # These two ways conflict with each other so we can only use one of them
+
+  # Using canned ACL will conflict with using grant ACL
+  acl = local.use_grant_acl ? null : "private"
+
+  # Use ACL policy grant to grant permissions to certain users, grant will conflict with canned ACL.
+  # We sometimes need to grant ceertain canonical users, 
+  # (eg. when it is a logging bucket, then we need to grant the awslogsdelivery account FULL_CONTROL permission
+  # more details of how to grant awslogsdelivery permissions in https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html)
+  # Details in how to find the canonical user id: https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html
+  dynamic "grant" {
+    for_each = [for pair in var.canonical_user_id_and_permissions : {
+      canonical_user_id = pair.canonical_user_id
+      permissions       = pair.permissions
+    }]
+
+    content {
+      id          = grant.value.canonical_user_id
+      permissions = grant.value.permissions
+      type        = "CanonicalUser"
+    }
+  }
 
   policy = data.aws_iam_policy_document.bucket_policy.json
 
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -1,5 +1,6 @@
 locals {
-  acl = length(var.grants) == 0 ? null : "private"
+  # If grants are defined, we use `grant` to grant permissions, otherwise it will use the `acl` to grant permissions
+  acl = length(var.grants) == 0 ? "private" : null
 
   tags = {
     project   = var.project
@@ -22,11 +23,14 @@ resource "aws_s3_bucket" "bucket" {
   # user FULL_CONTROL permission - https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html
   # How to find the canonical user id: https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html
   dynamic "grant" {
-    for_each = var.grants
+    for_each = [for pair in var.grants : {
+      canonical_user_id = pair.canonical_user_id
+      permissions       = pair.permissions
+    }]
 
     content {
-      id          = pair.canonical_user_id
-      permissions = pair.permissions
+      id          = grant.value.canonical_user_id
+      permissions = grant.value.permissions
       type        = "CanonicalUser"
     }
   }
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -1,5 +1,5 @@
 locals {
-  acl = length(var.canonical_user_id_and_permissions) == 0 ? null : "private"
+  acl = length(var.grants) == 0 ? null : "private"
 
   tags = {
     project   = var.project
@@ -22,7 +22,7 @@ resource "aws_s3_bucket" "bucket" {
   # user FULL_CONTROL permission - https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html
   # How to find the canonical user id: https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html
   dynamic "grant" {
-    for_each = var.canonical_user_id_and_permissions
+    for_each = var.grants
 
     content {
       id          = pair.canonical_user_id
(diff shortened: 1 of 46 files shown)


aws-s3-private-bucket/main.tf
```diff
@@ -1,4 +1,6 @@
 locals {
+  use_grant_acl = length(var.canonical_user_id_and_permissions) == 0 ? false : true
```
ryanking, line 2: I would rewrite as:

```suggestion
  acl = length(var.canonical_user_id_and_permissions) == 0 ? null : "private"
```

aws-s3-private-bucket/main.tf
```diff
@@ -10,7 +12,31 @@ locals {
...
 resource "aws_s3_bucket" "bucket" {
   bucket = var.bucket_name
-  acl    = "private"
+  # To define the access control list (ACL) of the bucket, there are two ways:
+  # 1. Define the canned ACL using the acl argument
+  # 2. Define the grant ACL using the grant argument
+  # These two ways conflict with each other so we can only use one of them
+
+  # Using canned ACL will conflict with using grant ACL
+  acl = local.use_grant_acl ? null : "private"
```
ryanking, line 21: ...and...

```suggestion
  acl = local.acl
```

aws-s3-private-bucket/main.tf
```diff
@@ -10,7 +12,31 @@ locals {
 
 resource "aws_s3_bucket" "bucket" {
   bucket = var.bucket_name
-  acl    = "private"
+  # To define the access control list (ACL) of the bucket, there are two ways:
+  # 1. Define the canned ACL using the acl argument
+  # 2. Define the grant ACL using the grant argument
+  # These two ways conflict with each other so we can only use one of them
```
ryanking, line 18:
```suggestion
# `grant` and `acl` conflict with each other - https://www.terraform.io/docs/providers/aws/r/s3_bucket.html#acl
```

aws-s3-private-bucket/main.tf
```diff
@@ -10,7 +12,31 @@ locals {
...
+    for_each = [for pair in var.canonical_user_id_and_permissions : {
+      canonical_user_id = pair.canonical_user_id
+      permissions       = pair.permissions
+    }]
+
+    content {
+      id          = grant.value.canonical_user_id
+      permissions = grant.value.permissions
+      type        = "CanonicalUser"
+    }
```
ryanking, line 38: I'm pretty sure this is valid, and if so it would be simpler: 

```suggestion
    for_each = var.canonical_user_id_and_permissions 

    content {
      id          = pair.canonical_user_id
      permissions =  pair.permissions
      type        = "CanonicalUser"
    }
```

but if I am wrong, nevermind.

aws-s3-private-bucket/variables.tf
```diff
@@ -62,3 +62,20 @@ variable public_access_block {
   type    = bool
   default = true
 }
+
+variable canonical_user_id_and_permissions {
+  type    = list(object({ canonical_user_id : string, permissions : list(string) }))
+  default = []
+  # A smaple of canonical_user_id_and_permissions variable is:
```
ryanking, line 69:
```suggestion
  # A sample of canonical_user_id_and_permissions variable is: 
```

ryanking, line 69: Can you put this example in the README instead?

aws-s3-private-bucket/main.tf
```diff
@@ -10,7 +12,31 @@ locals {
...
+  # These two ways conflict with each other so we can only use one of them
+
+  # Using canned ACL will conflict with using grant ACL
+  acl = local.use_grant_acl ? null : "private"
+
+  # Use ACL policy grant to grant permissions to certain users, grant will conflict with canned ACL.
+  # We sometimes need to grant ceertain canonical users, 
+  # (eg. when it is a logging bucket, then we need to grant the awslogsdelivery account FULL_CONTROL permission
+  # more details of how to grant awslogsdelivery permissions in https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html)
+  # Details in how to find the canonical user id: https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html
```
ryanking, line 27:
```suggestion
  # Use ACL policy grant to grant permissions to certain users.
  # Especially useful for logging buckets, where we  need to grant the awslogsdelivery
  # user FULL_CONTROL permission - https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html
  # How to find the canonical user id: https://docs.aws.amazon.com/general/latest/gr/acct-identifiers.html
```

aws-s3-private-bucket/variables.tf
```diff
@@ -62,3 +62,9 @@ variable public_access_block {
   type    = bool
   default = true
 }
+
+variable canonical_user_id_and_permissions {
```
ryanking, line 66: Can we rename to to just `grants`. The current naming does not seem to indicate what these users and permissions are used for.

aws-s3-private-bucket/variables.tf
```diff
@@ -62,3 +62,9 @@ variable public_access_block {
...
+  type        = list(object({ canonical_user_id : string, permissions : list(string) }))
+  default     = []
+  description = "A list of canonical user ID to permissions pairs. Used when we want to grant certain permissions to certain AWS accounts with canonical ID."
```
ryanking, line 69:
```suggestion
  description = "A list of canonical user ID to permissions pairs. Used when we want to grant permissions to AWS accounts via the S3 ACL system."
```

aws-s3-private-bucket/variables.tf
Dingying0410: sure! the fields for `canonical_user_id` and `permissions` already indicate the information of each value, using `grants` looks good to me. (updated the summary)

aws-s3-private-bucket/main.tf
Dingying0410: heads up the order was wrong, fixed in the latest commit

aws-s3-private-bucket/main.tf
Dingying0410: looks like we still need the original way, otherwise it complains `canonical_user_id` not declared

PR #206, opened 2020-06-24
[fix] [testing] make TDD more productive

ryanking:
This uses a technique inspired by [terratest's test-structure](https://github.com/gruntwork-io/terratest/tree/master/modules/test-structure) tools to make it easier to iterate
on this code.

The idea is that each test will be broken into stages 'options',
'apply', 'validate' and 'cleanup' which can be skipped but also persist
any state (like randomly generated options).

So you can do something like:

`make test TEST=./aws-s3-private-bucket SKIP=cleanup`

Which will

1. run 'options' to generate and persist a set of random variables
2. run 'apply' which will `terraform apply` to create resources and persist the state file
3. run 'validate' to make assertions about the resources created
4. skip 'cleanup'

This time will be dominated by how long it takes to apply the changes (seconds to ~1hr).

Then if you run:

`make test TEST=./aws-s3-private-bucket SKIP=cleanup,options,apply`

It will only run the validate step, which takes < 10s. You can re run
this, iterating on your tests until you are satisfied.

At some point you might realize you need to update the terraform code.
After you make those edits you can run:

`make test TEST=./aws-s3-private-bucket SKIP=cleanup,options`

which will

1. skip options and use the persisted options
2. run apply, using existing options and state file. This will mean that
   for some resources you can do an update-in-place, potentially much
   faster.
3. run validate
4. skip cleanup

When all done, you can run cleanup and then re-run with no skips to make
sure it works end-to-end.


aws-s3-private-bucket/main.tf
```diff
@@ -102,7 +102,10 @@ data "aws_iam_policy_document" "bucket_policy" {
   statement {
     sid       = "EnforceTLS"
     actions   = ["*"]
-    resources = ["arn:aws:s3:::${var.bucket_name}/*"]
+    resources = [
```
ryanking, line 105: Note that writing these tests led me to realize we had a bug here!

testutil/util.go
```diff
@@ -138,7 +138,10 @@ func Cleanup(t *testing.T, options *terraform.Options) {
 
 func ListEnvVar(name string) []string {
 	s := os.Getenv(name)
-	return strings.Split(s, ",")
+	splitFn := func(c rune) bool {
```
ryanking, line 141: without this change an input of "" will give a slice with {""}

PR #210, opened 2020-07-03
Enable logging with private s3 bucket

alldoami:
### Summary
[This](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html#AccessLogsBucketAndFileOwnership) doc describes that if you remove permissions for the `awslogsdelivery` account, CloudFront won't be able to save logs to the S3 bucket. Terraform is wiping this out every time we update s3 buckets. We have to explicitly create grants to `awslogsdelivery`. Inspired by: https://github.com/FB-PLP/terraform-infra-management/pull/386, creating a module for this so that it will be easier to configure.

### Test Plan
Updating a module that is referencing this module and seeing if logs show up (since they have disappeared when buckets were updated last). Tests in go.


aws-s3-private-bucket/main.tf
```diff
@@ -1,6 +1,8 @@
 locals {
   # If grants are defined, we use `grant` to grant permissions, otherwise it will use the `acl` to grant permissions
-  acl = length(var.grants) == 0 ? "private" : null
+  acl = length(var.grants) == 0 ? "private" : (
+    var.log_delivery_write_acl_enable ? "log-delivery-write" : null
```
jgadling, line 4: Think this needs a fix -- the log delivery flag will only work here if we have grants AND log_delivery_write_acl_enable set.

alldoami: redid, but had to put in a try because https://github.com/hashicorp/terraform/issues/25014

alldoami: I did find this: https://github.com/hashicorp/terraform/issues/22131#issuecomment-513020615, which is what I had previously.

aws-s3-private-bucket/variables.tf
```diff
@@ -68,3 +68,9 @@ variable grants {
   default     = []
   description = "A list of canonical user ID to permissions pairs. Used when we want to grant permissions to AWS accounts via the S3 ACL system."
 }
+
+variable log_delivery_write_acl_enable {
+  type        = bool
+  default     = true
```
jgadling, line 74: The default should probably be false?

jgadling:
Ok, this looks like a good start to me -- let's test it in the cellxgene dev environment before merging.

aws-s3-private-bucket/variables.tf
```diff
@@ -68,3 +68,9 @@ variable grants {
   default     = []
   description = "A list of canonical user ID to permissions pairs. Used when we want to grant permissions to AWS accounts via the S3 ACL system."
 }
+
+variable log_delivery_write_acl_enable {
+  type        = bool
+  default     = false
+  description = "Enables logging"
```
ryanking, line 75: Can we have a more descriptive description?

aws-cloudfront-logs-bucket/main.tf
```diff
@@ -0,0 +1,36 @@
+locals {
+  # Define the grant ACL for the Cloudfront logging S3 bucket,
+  # In order for the awslogsdelivery account to write log files to the bucket,
+  # we need to grant the AWS log delivery group the FULL_CONTROL access to the logging bucket
+  # LP's AWS account also has the FULL_CONTROL access to the bucket, this is specified by the canonical user id
```
jgadling, line 5: This comment is confusing -- please remove the reference to LP

PR #212, opened 2020-07-07
[feature] custom CORS rules for private S3 buckets

palasha:
As title.

My specific use case is that I want to allow presigned URL uploads into a private bucket. That requires a CORS configuration.

Test Plan: not sure how to do that


palasha:
I'm not sure how to test this. First it asked me to install terraform (which shouldn't I already have? I use it every day already via fogg) and then the `make lint` fails with `Error 3` but it doesn't tell me whats actually wrong?

palasha:
Try a different approach to avoid build failure

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/ad567203968e4918bd19ae3e99bf0e89a5cb8526..a7c60f6faa520cb3fc116db9c00e6932fbc4afa6).

palasha:
approach #3

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/9f09999df2b6504669d8e3e99e47c10082407eda..fbe4dc8c94ae69497f142fc6f2a596918bf16f27).

palasha:
copy lifecycle_rules by adding .value

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/0c8ede7b23814a7c05fbeadcb864e7ee85627cda..b14d9b5dd2f9b478dac4f295192589e8ac9f15f5).

palasha:
This worked when I made my terraform module point at the commit hash for this PR and then did make apply.

palasha:
add transfer acceleration as well

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/289c3f1a180279be60245c2d43d9f357344282a3..68c7728b42d34ee269b4fec56ed37ced242f0d84).

ryanking:
> I'm not sure how to test this. First it asked me to install terraform (which shouldn't I already have? I use it every day already via fogg)

Fogg does not install terraform globally, it uses tflint to install a copy in the infra repository (in the .fogg directory).

> and then the `make lint` fails with `Error 3` but it doesn't tell me whats actually wrong?

Seems you figured this out.

ryanking:
If you want to write an automated test, the tests for this module are here -https://github.com/chanzuckerberg/cztack/blob/master/aws-s3-private-bucket/module_test.go

You could either add configuration and assertions to the one test func that is there, or add a new test for this functionality.

ryanking:
(note that this testing technique is new, undocumented and a bit rough around the edges)

palasha:
> Seems you figured this out.

Sort of. I manually formatted the equal signs to line up since that's usually the problem, and then `make lint` started working? Either way, problem solved.

> If you want to write an automated test, the tests for this module are here -

I tried running the test locally, but I get oidc auth errors, so I think I don't have permission for the CZI accounts to do that. Took a quick glance at the go code for the test but I think its a bit too deep of a hole to go into at this particular moment.

PR #215, opened 2020-07-15
[refactor] use testutil.Test

ryanking:
Apply the pattern established with https://github.com/chanzuckerberg/cztack/pull/206.

This refactor is a small step toward making TDD-driven workflows more management for modules.

### Test Plan
* CI

### References
* https://github.com/chanzuckerberg/cztack/pull/206

### DIFF
--- a/aws-nuke.yml
+++ b/aws-nuke.yml
@@ -18,8 +18,10 @@ accounts:
           type: contains
           value: bastion
       EC2Instance:
-        - property: "tag:env"
-          value: "cztack-ci"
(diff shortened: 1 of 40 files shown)


aws-nuke.yml
```diff
@@ -18,8 +18,10 @@ accounts:
           type: contains
           value: bastion
       EC2Instance:
-        - property: "tag:env"
-          value: "cztack-ci"
+        # - property: "tag:env"
```
edulop91, line 21: why comment this? should it be deleted?

PR #218, opened 2020-07-20
[breaking] aws-lambda-function: create role

ryanking:
Create an IAM role for the function rather than taking as a variable.

The role needs special permissions to be useful, so best to encapsulate
those here. Also create a cloudwatch log group.

### Test Plan
* tests added

### References


aws-lambda-function/main.tf
```diff
@@ -38,3 +37,59 @@ resource aws_lambda_function lambda {
...
+EOF
+
+  tags = local.tags
+}
+
+resource aws_cloudwatch_log_group log {
+  name = "/aws/lambda/${local.name}"
+}
+
+resource aws_iam_policy lambda_logging {
```
jgadling, line 68: I'm guessing that users of this module are going to want to supply customized roles or policies

ryanking: We usually use the inverse pattern - return the role name and let the caller attach things to it. I have an example coming shortly.

PR #224, opened 2020-08-03
[feature] setting lambda cloudwatch log retention

ryanking:
Enable setting log retention parameter for the cloud watch log groups
attached to lambdas.

### Test Plan
* CI

### References

### DIFF
--- a/aws-lambda-function/main.tf
+++ b/aws-lambda-function/main.tf
@@ -62,7 +62,8 @@ EOF
 }
 
 resource aws_cloudwatch_log_group log {
-  name = "/aws/lambda/${local.name}"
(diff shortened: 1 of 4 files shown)


aws-lambda-function/main.tf
```diff
@@ -62,7 +62,8 @@ EOF
 }
 
 resource aws_cloudwatch_log_group log {
-  name = "/aws/lambda/${local.name}"
+  name              = "/aws/lambda/${local.name}"
```
edulop91, line 65: unrelated but: should the aws_lambda_function `depends_on` this `aws_cloudwatch_log_group` ?

ryanking: yeah probably

PR #233, opened 2020-09-25
[fix] s3 bucket abort_incomplete_multipart_upload_days should be read from lifecycle

Dingying0410:
### Test Plan
Tested in LP's terraform repo (PR 553)

### References

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -64,7 +64,7 @@ resource "aws_s3_bucket" "bucket" {
       prefix                                 = lookup(lifecycle_rule.value, "prefix", null)
       tags                                   = lookup(lifecycle_rule.value, "tags", null)
       enabled                                = lookup(lifecycle_rule.value, "enabled", false)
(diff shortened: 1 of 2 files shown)


mbarrien (changes requested):
This will break a lot of other existing usage of the standalone variable and leaves that standalone variable unused but still defined. If you're looking for something where the lifecycle value can optionally override the default, you can use var.abort_incomplete_multipart_upload_days as the default instead of 0, but for that we'd have to be sure that's the desired behavior that doesn't break other code.

Dingying0410:
> If you're looking for something where the lifecycle value can optionally override the default, you can use var.abort_incomplete_multipart_upload_days as the default instead of 0, 

yeah! this makes sense

> but for that we'd have to be sure that's the desired behavior that doesn't break other code.

Hmm I think the current way for either following the general `abort_incomplete_multipart_upload_days` or `abort_incomplete_multipart_upload_days` in each lifecycle rule would be safe (tested in my local branch), lmk if that address your concern :)

Thanks! @mbarrien

PR #235, opened 2020-09-28
[breaking] [fix] add block for bucket policy to fix the not unsynced change in importing the bucket

Dingying0410:
### Summary
the bucket policy when import: In the current  version, after I do an import the `module....aws_s3_bucket.bucket `  in when I wanted to import a bucket for LP, it will show that 
```
# module.....bucket will be updated in-place
  ~ resource "aws_s3_bucket" "bucket" {
      + acceleration_status         = "Suspended"
      + acl                         = "private"
        ....

      + force_destroy               = false
        .....
      + policy                      = jsonencode(
            {
              + Statement = [
                  + {
                      + Action    = "s3:DeleteObject"
                      + Effect    = "Allow"
                      + Principal = {
                          + Service = "s3.amazonaws.com"
                        }
                      + Resource  = "....."
                      + Sid       = ""
                    },
                  + {
                      + Action    = "*"
                      + Condition = {
                          + Bool = {
                              + aws:SecureTransport = "false"
                            }
                        }
                      + Effect    = "Deny"
                      + Principal = "*"
                      + Resource  = [
                          ......
                        ]
                      + Sid       = "EnforceTLS"
                    },
                ]
              + Version   = "2012-10-17"
            }
        )
        region                      = "us-east-1"
        request_payer               = "BucketOwner"
      ~ tags                        = {
          .....
        }

        .........

Plan: 0 to add, 1 to change, 0 to destroy.
```

this indicates that the policy was not imported with the bucket, and if I accidentally have an error in the policy variable (eg. the `principals` here should be "*" instead of `type: service`, it will not know it or reflect it, so it would be necessary to import the policy as well. 

**Changes in this PR**: 
- add the `aws_s3_bucket_policy` following `aws_iam_policy_document` and remove the `policy` in the `aws_s3_bucket`

(Other changes are in https://github.com/chanzuckerberg/cztack/pull/233, I want to separate the PR for different changes, but want to facilitate my testing in the LP repo)

### Test Plan
Tested in a PR in LP (PR 533)
- I can import the policy correctly,  it doesn't show the changes for adding policy as in the plan above
- if we have unsynced changes in the code with the  bucket policy, it will be detected in make plan


ryanking (approved):
I think this is fine and we should not worry about the failing test, since it fails at destroy time (presumably something related to the creation of the bucket is still happening at that point.

ryanking (approved):
It was a real issue! Just pushed a fix.

Dingying0410:
thanks for the fix @ryanking

PR #243, opened 2020-10-08
[feature] domain redirects via Cloudfront

palasha:
There should be a canonical way to redirect from a naked domain like
http://foo.com or https://foo.com to https://www.foo.com, while
including the appropriate security headers like HSTS.

This is a first pass at such an abstraction.

I separated out the security-headers lambda since it is independently
useful in other cloudfront distribution settings. It's not very
configurable yet, and doesn't quite live up to the promise of its name,
but perhaps will get there in time.

Test Plan:
push the PR so I can try using it in another repo.

PR #244, opened 2020-10-08
[feature] HSTS headers in aws-single-page-static-site

palasha:
This module already enforces that the site is accessed with HTTPS via
the cloudfront `"redirect-to-https"` policy.

Since that's true, we can also set HSTS headers on all the responses
using a cloudfront lambda, that, I've conveniently created already in
https://github.com/chanzuckerberg/cztack/pull/243

This does that.

Test Plan:
push up PR
update connections-infra module ref to new commit
make apply
curl to see if it works


palasha:
override function name to prevent collision with naked domain redirect in same component

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/30669aeb2b76cc5b5783f8d8a2c8a25739f2d561..ccb96a45b50691946da6a11d35e50977dfa8ee47).

aws-single-page-static-site/main.tf
```diff
@@ -82,6 +82,17 @@ resource "aws_s3_bucket_public_access_block" "bucket" {
   restrict_public_buckets = true
 }
 
+module "security_headers_lambda" {
+  source = "git@github.com:chanzuckerberg/cztack//aws-lambda-edge-add-security-headers?ref=7ba56765f2401062fe26bc601920046aa7a739c7"
```
ryanking, line 86:
```suggestion
  source = "../aws-lambda-edge-add-security-headers"
```

ryanking (changes requested):
1. we should reference the other cztack module via a relative path

2. do you think this is safe to include for all usage or this module? I think so but want to make sure that you have thought through it.

palasha:
relative path to cztack module

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/0482e73661830525f0add1e3a3285ef8e432a962..1ca8bc7624dbd2f7e73e2c70e58deb8550fa9750).

palasha:
> do you think this is safe to include for all usage or this module?

My thought was that since this module already enforces https, that this should be safe / better for existing users of the module as well.

damien (approved):
Reviewed this and the associated HSTS lambda. Well done!

ryanking (approved):

PR #257, opened 2020-12-02
Point ec2-poweruser at specific EC2 instances

michaelmcchen:
### Summary
In Learning Platform, we'd like to allow developers to be able to stop and start their dev servers via a separate role that they can assume via Okta (similar to `readonly-all-aws`). This PR makes `aws-iam-role-ec2-poweruser` more configurable, so we can specify the start/stop actions against EC2 instances in `us-west-2`. 

Since this module is called `aws-iam-role-ec2-poweruser`, I'm also open to creating a specific module within LP's terraform-infra-management repo instead of making this module configurable, to keep the "poweruser" semantics consistent (seems like CZI's definition of poweruser stems across all actions for all instances of a resource type in an account). 

### Test Plan
Utilized this module in https://github.com/FB-PLP/terraform-infra-management/pull/875 and the Terraform plan seems reasonable (see Test Plan in that PR)

### References


aws-iam-role-ec2-poweruser/variables.tf
```diff
@@ -25,6 +25,18 @@ variable "saml_idp_arn" {
   description = "The AWS SAML IDP arn to establish a trust relationship. Ignored if empty or not provided."
 }
 
+variable "iam_actions" {
+  type        = list(string)
+  default     = ["ec2:*"]
+  description = "List of EC2 API Actions that we are granting this role. "
```
edulop91, line 31: This could be somewhat "abused" to grant perms outside the ec2 scope. I wonder if we should consider forcing the ec2 namespace by doing something like `formatlist("ec2:%s", var.iam_actions)` and then the default action is `["*"]`

aws-iam-role-ec2-poweruser/variables.tf
```diff
@@ -25,6 +25,18 @@ variable "saml_idp_arn" {
   description = "The AWS SAML IDP arn to establish a trust relationship. Ignored if empty or not provided."
 }
 
+variable "iam_actions" {
```
edulop91, line 28: This is a bit of nit picking but I've been trying to move to a style where related variables are encoded into objects so we inherently express their "relatedness".

Maybe something like:
```
variable iam_ec2 {
  type = "object({actions: list(string), resources: list(string)})
  description = "....."
}
```

thomsonyeh-czi (approved):

aws-iam-role-ec2-poweruser/variables.tf
michaelmcchen: Ah yeah, that's a good point. I'll add this

edulop91 (approved):

aws-iam-role-ec2-poweruser/module_test.go
```diff
@@ -8,19 +8,25 @@ import (
...
+
 			return tftest.Options(
 				tftest.IAMRegion,
 				map[string]interface{}{
 					"role_name":         random.UniqueId(),
 					"source_account_id": curAcct,
-					"tags": map[string]string{
-						"test": random.UniqueId(),
-					},
+					"project": project,
```
cziie, line 26: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>File is not `gofmt`-ed with `-s` (gofmt)

PR #260, opened 2020-12-04
add tags specified by tftest

michaelmcchen:
### Summary
My previous PR was failing because https://github.com/chanzuckerberg/cztack/commit/209f50326c28c9fd05d1a5b5e1a6ac6acdead1fd brought in https://github.com/chanzuckerberg/go-misc/commit/d32d5af6473ec4b5f4e0ad01fea82066a5ff149f which enforces the env/owner/service/project tagging scheme over a `tags` variable. This fixes it for the other modules (it seems the test suite is only configured to run certain module tests when there are changes to those modules), which is why this didn't get picked up earlier. 

### Test Plan
I don't have permissions to run the full test suite, but `make test-ci TEST=./MODULE` seems to make it past the initial Terraform static checks. Let's see if CI tests fully pass. 

### References

### DIFF
--- a/aws-iam-role-bless/main.tf
+++ b/aws-iam-role-bless/main.tf
@@ -19,5 +19,8 @@ module "client" {
   iam_path           = var.iam_path
   source_account_id  = var.source_account_id
   source_account_ids = var.source_account_ids
-  tags               = var.tags
+  env                = var.env
+  owner              = var.owner
+  service            = var.service
+  project            = var.project
 }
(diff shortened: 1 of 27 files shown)


kuannie1 (approved):

kuannie1:
Hi @michaelmcchen 
I struggled with this problem too! I think you're taking the right approach.

Do you have access to the `cztack-ci-1` account? If you do, you can run `make test TEST=./MODULE` locally.

PR #261, opened 2020-12-04
[fix] snowflake-table-grant-all: Also allow specifying specific priv grants

edulop91:
### Summary
Augment the table grant ALL with the ability to issue individual privs to extra roles/shares. 

### Test Plan
just terraform init for now

### References

### DIFF
--- a/aws-iam-role-bless/README.md
+++ b/aws-iam-role-bless/README.md
@@ -37,11 +37,14 @@ output "..." {
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | bless\_lambda\_arns | List of bless lambda arns | `list` | n/a | yes |
+| env | Env for tagging and naming. See [doc](../README.md#consistent-tagging) | `string` | n/a | yes |
 | iam\_path | IAM path | `string` | `"/"` | no |
+| owner | Owner for tagging and naming. See [doc](../README.md#consistent-tagging) | `string` | n/a | yes |
+| project | Project for tagging and naming. See [doc](../README.md#consistent-tagging) | `string` | n/a | yes |
 | role\_name | The name for the role | `string` | n/a | yes |
+| service | Service for tagging and naming. See [doc](../README.md#consistent-tagging) | `string` | n/a | yes |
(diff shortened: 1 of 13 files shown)


aws-iam-role-bless/README.md
```diff
@@ -37,11 +37,14 @@ output "..." {
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | bless\_lambda\_arns | List of bless lambda arns | `list` | n/a | yes |
+| env | Env for tagging and naming. See [doc](../README.md#consistent-tagging) | `string` | n/a | yes |
```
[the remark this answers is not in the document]
edulop91: not unused actually

snowflake-table-grant-all/variables.tf
```diff
@@ -49,3 +49,14 @@ variable with_grant_option {
 
   default = null
 }
+
+variable per_privilege_grants {
+  type        = map(object({ roles : list(string), shares : list(string) }))
+  description = <<EOF
+  A map of privileges to authorized roles and shares. Privileges must be UPPER case.
```
alldoami, line 56: To prevent errors when the privs are in lower case, is there a way for us to convert privs to uppercase through the provider?

ryanking: why not convert them to upper in terraform?

edulop91: I think the provider is already case-insensitive. The issue here is more with looking up a key in this map being case-sensitive.

alldoami: Do `make plan`s fail when the privs are lowercase? Or you only see that you did the privs wrong in the apply?

edulop91: they'd be ignored

edulop91: >why not convert them to upper in terraform?
I could do something like that:
`upper_per_privilege_grants = {for priv, grant in var.per_privilege_grants: upper(priv) => grant }`
But don't think it would be well-defined for all inputs.

ryanking (approved):

PR #263, opened 2020-12-08
aws-cloudwatch-logs-retention-manager

palasha:
Create a scheduled lambda function to iterate through all Cloudwatch Log
Groups and enforce a maximum retention. This is particularly useful for
deletion efforts on the Education team, because Lambda@Edge functions
(amongst others) can log PII, but have no way to control retention via
Terraform.

It's a good question whether we want this to be:
- Opt-out: Applies to all log groups unless specified otherwise.
- Opt-in: Only applies to specific prefixes.

I think there's probably not much downside in the opt-out approach. It's
the safest in terms of deletion requirements, and its rare for us to have logs
that we would consider "permanent", since any of them could potentially
contain PII.

Test Plan: Dry runs in the console with the `put_retention_policy`
commented out.

### DIFF
--- a/aws-cloudwatch-log-retention-manager/main.tf
+++ b/aws-cloudwatch-log-retention-manager/main.tf
@@ -1,0 +1,117 @@
+locals {
+  tags = {
+    managedBy = "terraform"
+    env       = var.env
+    project   = var.project
+    service   = var.service
+    owner     = var.owner
+  }
+
+  lambda_name = "${var.project}-${var.env}-${var.service}-cloudwatch-retention"
+}
+
+data archive_file lambda {
+  type        = "zip"
+  output_path = "${path.module}/build/lambda.zip"
+  source {
+    filename = "function.rb"
+    content  = <<-EOF
+      require 'aws-sdk-ec2'
+      require 'aws-sdk-cloudwatchlogs'
+      MAXIMUM_RETENTION = ${var.maximum_retention}
+
+      def handler(event:, context:)
+        regions = Aws::EC2::Client.new.describe_regions.regions.map(&:region_name)
+        regions.each do |region|
+          puts "Processing Region: #{region}"
+          cloudwatch = Aws::CloudWatchLogs::Client.new(region: region)
+          cloudwatch.describe_log_groups.each do |log_group_batch|
+            log_group_batch.log_groups.each do |log_group|
+              name = log_group.log_group_name
+              retention = log_group.retention_in_days
+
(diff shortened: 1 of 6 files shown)


palasha:
- fix bugs by testing with a local copy of the module
- fixup readme and add to CI
- uncomment the dangerous parts (the actual retention update call)

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/358f9ad8cbad76617d07daa78d19ccffb1f1a01a..46953c964c6bfec5c4cbd3ddd91ba21c2f1fa750).

aws-cloudwatch-log-retention-manager/main.tf
```diff
@@ -0,0 +1,117 @@
...
+    env       = var.env
+    project   = var.project
+    service   = var.service
+    owner     = var.owner
+  }
+
+  lambda_name = "${var.project}-${var.env}-${var.service}-cloudwatch-retention"
+}
+
+data archive_file lambda {
```
cziie, line 13: ⚠️ **[tflint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing version constraint for provider "archive" in "required_providers"

edulop91:
@damien ^ is probably relevant wrt the custodian work

damien:
@palasha This is interesting work! I don't see anything wrong with this module per sey, but I'd love to see the edge functions your team is using. Questions I have:

1. What kind of data is being leaked by the Lambda function(s) in question?
2. Can this data leakage be avoided?

We (InfraSec) would be very interested to see where this kind of friction is happening ; it'll better inform us on what needs to change and what your team's specific needs are. :)

damien (approved):

palasha:
> I'd love to see the edge functions your team is using

We have quite a few of them, all associated with Cloudfront distributions. Some do authentication for video and image distribution, others add security headers. In all cases the logging is really helpful for debugging generally, and auditing in the case of a security incident, but then often also contains IP addresses or a user ID. We could probably change to not having logging, but its then much harder to build and debug these functions when they break.

palasha:
- pin archive provider version
- update readme example
- cleanup some naming in lambda code

:mag_right: [Click to see changes since last update](https://github.com/chanzuckerberg/cztack/compare/e5e2f72eddea7cc752d853247a8e707f650570cb..c3b50d81b8bdd155174d7a54bcaa40c17ed6d992).

PR #278, opened 2021-01-12
[feature] aws-s3-public-bucket require https, allow disabling versioning

mbarrien:
### Summary
Adds functionality from CZI's private version of this module.

* Add ability to disable versioning by default on public S3 buckets.
  Leaves default behavior of enabling versioning intact
* Optionally require TLS for every get object. Default to true.

### DIFF
--- a/aws-s3-public-bucket/main.tf
+++ b/aws-s3-public-bucket/main.tf
@@ -37,22 +37,25 @@ data "aws_iam_policy_document" "bucket_policy" {
   # Deny access to bucket if it's not accessed through HTTPS
   source_json = var.bucket_policy
 
-  statement {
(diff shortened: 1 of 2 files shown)


ryanking (approved):

ryanking (changes requested):
I don't think we can require TLS unconditionally because S3 website hosting still only works with http. :(

mbarrien:
This is a separate module from aws-single-page-static-site. In the only context it is used in CZI's code base, this has been working fine.

mbarrien:
Made TLS optional via another variable. Please re-review.

kuannie1 (approved):

PR #279, opened 2021-01-20
enable clouldwatch log exports for aurora postgresql

srm78:
Enable logs for aurora postgresql


aws-aurora-postgres/main.tf
```diff
@@ -26,6 +26,7 @@ module "aurora" {
   rds_cluster_parameters              = var.rds_cluster_parameters
   iam_database_authentication_enabled = var.iam_database_authentication_enabled
   performance_insights_enabled        = var.performance_insights_enabled
+  enabled_cloudwatch_logs_exports     = ["audit", "error", "general", "slowquery", "postgresql"]
```
Bento007, line 29: This works for me, but would it be better to make it something that can be optional for other users?

ryanking: agreed

ryanking (approved):

PR #283, opened 2021-02-01
[fix] ecs role deprecation

ryanking:
AmazonEC2ContainerServiceFullAccess is deprecated and can no longer be attached.

### Test Plan
* ci

### References
* https://docs.aws.amazon.com/AmazonECS/latest/developerguide/managed-policy-transition.html


edulop91:
@ryanking looks like an empty changeset

ryanking:
It looks like this accidentally got included in #281

PR #302, opened 2021-04-13
Initial updates for supporting s3 bucket logging targets.

jgadling:
### Summary
Add support for logging S3 access to another bucket or prefix to the private bucket module.

### Test Plan
I generated a tf file that created two buckets (a content bucket, and a logging bucket) and set the logging config of one bucket to refer to the other. The logging config in the plan had logging enabled for the content bucket and disabled for the logging bucket.

Tests pass.

### References

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -110,11 +110,13 @@ resource "aws_s3_bucket" "bucket" {
     }
   }
 
-  # TODO
-  #   logging {
-  #   target_bucket = ""
-  #   target_prefix = ""
-  # }
+  dynamic "logging" {
+    for_each = var.logging_bucket == null ? [] : [var.logging_bucket]
+    content {
+      target_bucket = var.logging_bucket.name
+      target_prefix = var.logging_bucket.prefix
+    }
+  }
 
   server_side_encryption_configuration {
     rule {
--- a/aws-s3-private-bucket/variables.tf
+++ b/aws-s3-private-bucket/variables.tf
@@ -71,8 +71,8 @@ variable "transfer_acceleration" {
   default = false
 }
 
-variable logging_bucket {
-  type        = object({name = string, prefix = string})
+variable "logging_bucket" {
+  type        = object({ name = string, prefix = string })
   description = "Log bucket name and prefix to enable logs for this bucket"
   default     = null
 }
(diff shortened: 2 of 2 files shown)


aws-s3-private-bucket/main.tf
```diff
@@ -110,11 +110,13 @@ resource "aws_s3_bucket" "bucket" {
...
-  #   target_bucket = ""
-  #   target_prefix = ""
-  # }
+  dynamic "logging" {
+    for_each = var.logging_bucket == null ? [] : [var.logging_bucket]
+    content {
+      target_bucket = var.logging_bucket.name
+      target_prefix = var.logging_bucket.prefix
+    }
+  }
```
edulop91, line 119: do you know what's the difference between this and cloudtrail logs?

aws-s3-private-bucket/main.tf
```diff
@@ -110,11 +110,13 @@ resource "aws_s3_bucket" "bucket" {
...
 
-  # TODO
-  #   logging {
-  #   target_bucket = ""
-  #   target_prefix = ""
-  # }
+  dynamic "logging" {
+    for_each = var.logging_bucket == null ? [] : [var.logging_bucket]
+    content {
+      target_bucket = var.logging_bucket.name
```
mbarrien, line 116: Don't know if it's better to do each.name/each.prefix, or leave it like this.

aws-s3-private-bucket/main.tf
jgadling: Here's a pretty interesting comparison: https://www.netskope.com/blog/aws-s3-logjam-server-access-logging-vs-object-level-logging

It looks like cloudtrail is more complete and more reliable, but also more complicated to configure.

maniarathi: There's a table in here that I used to determine that S3 logs was the way forward: https://docs.aws.amazon.com/AmazonS3/latest/userguide/logging-with-S3.html. Ultimately it was a combination of ease to set up (took me just a few minutes to set up and start querying) plus the fact that the S3 logs contain more useful information for us (i.e. Object Size, Total Time, Turn-Around Time, and HTTP Referer).

maniarathi (approved):

aws-s3-private-bucket/main.tf
jgadling: yeah, I could go either way as well. I think the each syntax is a little bit more confusing though for an iterator where we'll always have 0 or 1 loops.

mbarrien (approved):

PR #309, opened 2021-04-23
Fix running aws cli on GH Actions/Azure, get modules dynamically

mbarrien:
* Fixes an issue where running the aws cli (specifically aws sts get-caller-identity) would exit with error 255, and fail the build. It turns out that since we use ubuntu-latest (which recently upgraded to ubuntu 20.04 with aws cli v2) we were running into <https://github.com/aws/aws-cli/issues/5262>. When run on a machine that is running on Azure cloud (as Github Actions does), since the Azure metadata service runs on the same 169.254.169.254 location that AWS metadata service does. We disable using the metadata service by globally setting AWS_EC2_METADATA_DISABLED environment variable.
* Instead of having enumerate every single module and have to remember to change .github/workflows/ci.yml to add new modules/remove deprecated modules, we use a feature added by Github Actions in April 2020 (https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/) to take in the matrix via a JSON variable, and calculate the modules list dynamically. You can see this working in some previous Github Actions CI runs in this repo while I was testing this.
* Fixes tests that were committed in a broken state. Mostly cases where the module requires project/env/service/owner inputs but are not using tftest.Options which provides them, or the opposite case of modules that don't need those fields but used tftest.Options which thus gave extra arguments.

### DIFF
--- a/aws-iam-policy-cwlogs/module_test.go
+++ b/aws-iam-policy-cwlogs/module_test.go
@@ -6,19 +6,24 @@ import (
 
 	"github.com/chanzuckerberg/go-misc/tftest"
 	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/gruntwork-io/terratest/modules/terraform"
 )
 
 func TestAWSIAMPolicyCwlogs(t *testing.T) {
 	roleName := tftest.CreateRole(t)
 	defer tftest.DeleteRole(t, roleName) //nolint
 
-	terraformOptions := tftest.Options(
-		tftest.IAMRegion,
-		map[string]interface{}{
+	terraformOptions := &terraform.Options{
+		TerraformDir: ".",
+
+		EnvVars: map[string]string{
+			"AWS_DEFAULT_REGION": tftest.IAMRegion,
+		},
+		Vars: map[string]interface{}{
 			"role_name": roleName,
 			"iam_path":  fmt.Sprintf("/%s/", random.UniqueId()),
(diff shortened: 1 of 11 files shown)


aws-iam-policy-cwlogs/module_test.go
```diff
@@ -12,13 +12,17 @@ func TestAWSIAMPolicyCwlogs(t *testing.T) {
 	roleName := tftest.CreateRole(t)
 	defer tftest.DeleteRole(t, roleName) //nolint
 
-	terraformOptions := tftest.Options(
-		tftest.IAMRegion,
-		map[string]interface{}{
+	terraformOptions := &terraform.Options{
```
cziie, line 15: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>undeclared name: `terraform` (typecheck)

aws-sns-lambda/module_test.go
```diff
@@ -8,16 +8,17 @@ import (
 )
 
 func TestAwsSnsLambda(t *testing.T) {
```
cziie, line 10: 🚫 **[golangci]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>unnecessary leading newline (whitespace)

edulop91 (approved):

PR #326, opened 2021-07-23
[fix] remove ssm:DescribeParameters permission from SSM reader policy

kuannie1:
When trying to use the `ssm:DescribeParameters` permission, I keep getting this warning: 
```
The actions in your policy do not support resource-level permissions and require you to choose All resources
```
The warning is right--I wasn't able to run `ssm:DescribeParameters` for a specific subpath in SSM.

### DIFF
--- a/aws-params-reader-policy/main.tf
+++ b/aws-params-reader-policy/main.tf
@@ -29,14 +29,15 @@ data "aws_iam_policy_document" "policy" {
     actions   = ["kms:Decrypt"]
     resources = [data.aws_kms_alias.parameter_store_key.target_key_arn]
   }
-  # To use this reader policy with chamber, we need to give blanket ssm:DescribeParameters perms
+  # Note (aku): To use this reader policy with chamber, we need to give blanket ssm:DescribeParameters perms
   # https://github.com/segmentio/chamber/blob/53baa2b77bb1ed279bea346533d50f4fb1c01ed1/store/ssmstore.go#L214-L217
-  # This should be okay because this is a list option and the permission won't allow you to read the contents
-  # Otherwise, we'd get this error: The actions in your policy do not support resource-level permissions and require you to choose All resources
+  # This should be okay because this is a list option and the permission won't allow you to read the secret values
+  # Without this, we'd get this error: The actions in your policy do not support resource-level permissions and require you to choose All resources
   statement {
     sid = "ChamberSSMReadRequirement"
     actions   = ["ssm:DescribeParameters"]
     resources = ["arn:aws:ssm:${var.region}:${data.aws_caller_identity.current.account_id}:parameter/*"]
+    
   }
 }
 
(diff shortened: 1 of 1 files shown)


aws-params-reader-policy/main.tf
```diff
@@ -30,6 +29,9 @@ data "aws_iam_policy_document" "policy" {
     actions   = ["kms:Decrypt"]
     resources = [data.aws_kms_alias.parameter_store_key.target_key_arn]
   }
+  statement {
+    actions = ["ssm:DescribeParameters"]
+  }
 }
 
 resource "aws_iam_role_policy" "policy" {
```
cziie, line 37: ⚠️ **[tflint]** <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Missing version constraint for provider "aws" in "required_providers"

aws-params-reader-policy/main.tf
```diff
@@ -30,6 +29,10 @@ data "aws_iam_policy_document" "policy" {
     actions   = ["kms:Decrypt"]
     resources = [data.aws_kms_alias.parameter_store_key.target_key_arn]
   }
+  statement {
+    actions = ["ssm:DescribeParameters"]
+    resources = ["arn:aws:ssm:${var.region}:${data.aws_caller_identity.current.account_id}:parameter/*"]
+  }
```
kuannie1, line 35: DescribeParameters doesn't require a Resource value. This was a relief because we can still limit the reach of other perms, like `ssm:GetParameters`, `ssm:GetParametersByPath`, `ssm:GetParameter`, and `ssm:GetParameterHistory`.

aws-params-reader-policy/main.tf
```diff
@@ -30,6 +29,10 @@ data "aws_iam_policy_document" "policy" {
     actions   = ["kms:Decrypt"]
     resources = [data.aws_kms_alias.parameter_store_key.target_key_arn]
   }
+  statement {
+    actions   = ["ssm:DescribeParameters"]
+    resources = ["arn:aws:ssm:${var.region}:${data.aws_caller_identity.current.account_id}:parameter/*"]
```
jacobrheath, line 34: I briefly remember working on this with you. Was there no way to make this resource more specific? I remember we were having trouble getting this working with the `param_resources` because it was too specific for some reason. This would allow you to list all the parameters, which seems like too much.

jacobrheath (changes requested):

aws-params-reader-policy/main.tf
jacobrheath: After doing some more digging, I think this is fine since describe parameters is a list option and the permission won't allow you to read the contents. According to the AWS console, its not even allowed to set a specific ARN with the following error: "The actions in your policy do not support resource-level permissions and require you to choose All resources". Can we document this in the statement so that this information is preserved?

jacobrheath (approved):

aws-params-reader-policy/main.tf
kuannie1: Noted! added a comment and an `sid` value

PR #320, opened 2021-06-29
[feature] assume role policy module

kuannie1:
Main changes:
* Create aws-assume-role-policy module
* replaced assume policy documents with this module output
* README changes

PR #329, opened 2021-07-30
[breaking] deprecate source account ID input for its plural var

kuannie1:
This variable duplication confused me while I was debugging PR #320 . I think we're ready to deprecate it.

This PR changes 3 things:
* TF Module definition
* golang setup tests
* README.md files

### DIFF
--- a/aws-iam-role-bless/variables.tf
+++ b/aws-iam-role-bless/variables.tf
@@ -3,11 +3,6 @@ variable "role_name" {
   description = "The name for the role"
 }
 
-variable "source_account_id" {
-  type        = string
-  description = "The source aws account id to allow sts:AssumeRole. DEPRECATED: Please use source_account_ids"
-}
-
 variable "source_account_ids" {
   type        = set(string)
   default     = []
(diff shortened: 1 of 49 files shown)


aws-iam-role-bless/variables.tf
```diff
@@ -3,11 +3,6 @@ variable "role_name" {
   description = "The name for the role"
 }
 
-variable "source_account_id" {
```
kuannie1, line 6: This is the main change--deleting blocks like this

aws-iam-role-infraci/variables.tf
```diff
@@ -12,10 +12,10 @@ variable "terraform_state_lock_dynamodb_arns" {
   description = "ARNs of the state file DynamoDB tables"
 }
 
-variable "source_account_id" {
+variable "saml_idp_arn" {
```
edulop91, line 15: while you are here, i wonder if worth making these like the oidc blocks so you can configure multiple saml idps

edulop91: (could be in a separate PR)

edulop91 (approved):

edulop91 (approved):

PR #340, opened 2021-09-13
feat[service arn]: adding service ARN to module outputs

jacobrheath:
This PR adds an output of the service's ARN to the ECS fargate module.

### DIFF
--- a/aws-ecs-service-fargate/outputs.tf
+++ b/aws-ecs-service-fargate/outputs.tf
@@ -35,8 +35,8 @@ output "alb_http_listener_arn" {
   description = "ALB HTTP listener ARN, only if HTTPS forwarding is disabled"
   value       = var.disable_http_redirect ? aws_lb_listener.http.arn : null
 }
-  
+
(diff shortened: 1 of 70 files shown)


aws-ecs-service-fargate/outputs.tf
```diff
@@ -34,4 +34,9 @@ output "alb_https_listener_arn" {
 output "alb_http_listener_arn" {
   description = "ALB HTTP listener ARN, only if HTTPS forwarding is disabled"
   value       = var.disable_http_redirect ? aws_lb_listener.http.arn : null
-}
\ No newline at end of file
+}
+
+output "ecs_service_arn" {
+  description = "The ARN of the ECS service that is created"
+  value       = aws_ecs_service.id
```
edulop91, line 41: this is probably missing something like `aws_ecs_service.<some name>.id`

jacobrheath: o duh

edulop91 (approved):

PR #341, opened 2021-09-13
fix[fargate service name]: misspelled name

jacobrheath:
This PR fixes the ECS service output names


aws-ecs-service-fargate/outputs.tf
```diff
@@ -38,5 +38,5 @@ output "alb_http_listener_arn" {
 
 output "ecs_service_arn" {
   description = "The ARN of the ECS unmanaged service that is created"
-  value       = var.manage_task_definition ? aws_ecs_service.managed[0].id : aws_ecs_service.unmanaged[0].id
+  value       = var.manage_task_definition ? aws_ecs_service.job[0].id : aws_ecs_service.unmanaged-job[0].id
```
kuannie1, line 41: hm, can you create intuitive `local` variables for "aws_ecs_service.job[0].id" and "aws_ecs_service.unmanaged-job[0].id" in the main.tf file, then use those intuitive variable names here? No idea what the `[0]` means.

jacobrheath: It's because the `aws_ecs_service` resources are created with a `count`, so the resource name itself is a list and I have to index it as a list. I don't think I will be able to create a variable to reference those.

kuannie1 (approved):

PR #400, opened 2022-06-06
fix: Update Snowflake Terraform provider source to Snowflake-Labs

jjb007:
### Summary
With the source of the Snowflake Terraform provider transitioned from CZI to Snowflake, hard-coded mentions need to be updated accordingly.

Also fixed code to validate CI configuration from YAML to handle string scalars as well as array for `$.jobs.test.strategy.matrix.module`.

### Test Plan
```
cd scripts/snowflake_generate_grant_all
make run
```

### References
[ONCALL-65]

[ONCALL-65]: https://czi-eng.atlassian.net/browse/ONCALL-65?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jakeyheath:
I ran `go mod tidy` to update the deps properly. Let's see if that fixes some of the test failures.

alldoami (approved):

jakeyheath (approved):

scripts/snowflake_generate_grant_all/ensure_ci.go
```diff
@@ -35,8 +35,12 @@ func ensureCI(names []string) error {
...
 
 	modules := sets.NewStringSet().Add(names...)
-	for _, module := range matrix["module"].([]interface{}) {
-		modules.Add(module.(string))
+	if _, ok := matrix["module"].([]interface{}); ok {
+		for _, module := range matrix["module"].([]interface{}) {
+			modules.Add(module.(string))
+		}
+	} else {
+		modules.Add(matrix["module"].(string))
```
jjb007, line 43: This is the code change I made. Let me know if this is appropriate/idiomatic for Go and/or the program itself.

jjb007 (commented):
Also, is it concerning that running `scripts/snowflake_generate_grant_all` generated all these other modules?

scripts/snowflake_generate_grant_all/ensure_ci.go
jakeyheath: I think the infra team back then was really into code generation so I feel like this not surprising. I think the idiomatic way of doing this parsing though would be to make a type for the github action yaml and call something like     `err = yaml.Unmarshal(yamlFile, &config)` instead of doing type casting. However, I am not sure it is worth the half an hour to do this.

jakeyheath: In short, I think this is fine and good 👍🏻

alldoami:
yayyy tests are passing!

PR #407, opened 2022-07-06
feat: Attach ssm permissions to ecs fargate job module

tapanr:
Adding an option to attach ssm permission to ecs fargate job module. This option can be used when the task running in ECS Fargate needs to use parameter store to retrieve secret credentials.

Background on why we needed this: We plan to deploy a docker image on ECS Fargate which needs to set some authentication credentials as its environment variables. Since these are confidential credentials, we have stored them in parameter store and hence ECS task needs access to that store to retrieve them.

### DIFF
--- a/aws-ecs-job-fargate/iam.tf
+++ b/aws-ecs-job-fargate/iam.tf
@@ -43,24 +43,24 @@ resource "aws_iam_role_policy" "task_execution_role_secretsmanager" {
 }
 
 data "aws_iam_policy_document" "ssm_permission" {
-  count = var.ssm_arn != null ? 1 : 0
+  count = var.ssm_parameter_store_arns != null ? 1 : 0
 
   statement {
     actions = [
       "ssm:GetParameters",
     ]
 
-    resources = [var.ssm_arn]
+    resources = var.ssm_parameter_store_arns
   }
 }
 
 resource "aws_iam_policy" "task_execution_role_ssm_permission" {
-  count  = var.ssm_arn != null ? 1 : 0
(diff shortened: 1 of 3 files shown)


aws-ecs-job-fargate/variables.tf
```diff
@@ -101,3 +101,9 @@ variable "ordered_placement_strategy" {
   default     = []
   description = "Placement strategy for the task instances."
 }
+
+variable "ssm_arn" {
+  type        = string
+  default     = null
+  description = "Parameter Store ARN"
```
michaelmcchen, line 108:
```suggestion
  description = "SSM Parameter Store ARN. If present, allows ECS task to make ssm:GetParameters call."
```

aws-ecs-job-fargate/iam.tf
```diff
@@ -41,3 +41,26 @@ resource "aws_iam_role_policy" "task_execution_role_secretsmanager" {
...
+
+data "aws_iam_policy_document" "ssm_permission" {
+  count = var.ssm_arn != null ? 1 : 0
+
+  statement {
+    actions = [
+      "ssm:GetParameters",
+    ]
+
+    resources = [var.ssm_arn]
```
michaelmcchen, line 53:
```suggestion
    resources = var.ssm_parameter_store_arns
```
I'd make this a list so that we're granting access to specific keys in Parameter Store. Even though you can somewhat achieve the same behavior by providing an ARN prefix, I think this is much more flexible.

alldoami (approved):
Thanks for reviewing @michaelmcchen! @tapanr lmk if you need a release after this

PR #417, opened 2022-08-30
feat: Add ability to pass size of ephemeral storage to ECS

lamroger:
### Summary
Add the ability to specify ephemeral_storage_size to ECS Fargate tasks. 

### Test Plan
Going to test in dev along-infra


lamroger:
Is it ok if i leave out a default value for `ephemeral_storage_size`? Will the property just be omitted if it's optional?

lamroger:
I made the default 21GiB which is greater than not providing the value (20 GiB). Please lmk if there's a way to conditionally add the block.

I ran `aws ecs describe-task-definition --task-definition connections-dev-thorn-safer:31 --profile czi-connections` which shows the value. The ECS console hides it for some reason.

Screen Shot 2022-08-30 at 5 16 26 PM

alexlokshin-czi (approved):

PR #454, opened 2023-02-15
feat: configure log retention in aws-lambda-edge-add-security-headers module

michaelmcchen:
### Summary
Make this parameter configurable for a consuming TF project. Otherwise it defaults to `null`, which is the equivalent of 0. A 0 day retention policy makes those logs essentially useless.

### Test Plan
I think somebody previously made a change in the UI to configure a 30 day retention policy on these logs. Hence, a plan (without this PR) was prompting TF to change the retention policy from 30 to 0 days. 

I tested this PR out by running `make plan` on the consuming TF project and noting that the 30 to 0 day change is no longer prompted.

### References


alexlokshin-czi (approved):

palasha:
> I think somebody previously made a change in the UI to configure a 30 day retention policy on these logs.

https://github.com/chanzuckerberg/cztack/blob/main/aws-cloudwatch-log-retention-manager/main.tf#L36

It might have been this? I dont know if the null translates into 0 when that script reads the configuration for the log group

PR #461, opened 2023-03-03
Allow force_destroy option on s3 buckets

katyho:
### Summary
This option is helpful when we want to destroy a bucket and delete all objects within it: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#force_destroy
### Test Plan

### References

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -24,7 +24,8 @@ locals {
 }
 
 resource "aws_s3_bucket" "bucket" {
-  bucket = var.bucket_name
+  bucket        = var.bucket_name
+  force_destroy = var.force_destroy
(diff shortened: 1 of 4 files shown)


jakeyheath (approved):

jakeyheath:
Funny I needed this same exact thing today on ECRs!

jakeyheath (dismissed):
Can you update the versions.tf file to make sure that the module is at a minimum of where the AWS provider added that feature. I think we have some old AWS providers out there that won't be able to pick up this latest change.

katyho:
The force_destroy option is quite old but found it was added here: https://github.com/hashicorp/terraform-provider-aws/blob/release/2.x/CHANGELOG.md#2290-september-20-2019

PR #475, opened 2023-04-11
feat: Update engine version and instance class defaults for aurora psql

alldoami:
instance class `db.r4.large` is part of previous generation classes now and we should be using a current type.


aws-aurora-postgres/variables.tf
```diff
@@ -112,7 +112,7 @@ variable "kms_key_id" {
 variable "engine_version" {
   type        = string
   description = "The version of Postgres to use. This should be a *prefix* if auto version upgrades are enabled. (Docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#engine_version)"
-  default     = "10"
+  default     = "15.2"
```
alexlokshin-czi, line 115: This could be a breaking change... Isn't it safer to upgrade one major version at a time?

PR #491, opened 2023-05-10
feat: output additional info from aws-aurora-postgres

katyho:
### Summary
Propagate additional outputs to aws-aurora-postgres module.

- cluster_resource_id is useful for if you want to enable iam authentication on the database. The IAM policy requires a reference to the resource id. https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html
- the cluster_id is useful for providing unique names for other resources that are tied to the aurora instance

### Test Plan

### References

### DIFF
--- a/aws-aurora-postgres/outputs.tf
+++ b/aws-aurora-postgres/outputs.tf
@@ -25,3 +25,7 @@ output "port" {
 output "cluster_resource_id" {
   value = module.aurora.cluster_resource_id
 }
+
+output "cluster_id" {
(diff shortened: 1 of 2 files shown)


alldoami (approved):
up to you but we can also just export `module.aurora` so that you can get all the attributes.

katyho:
> up to you but we can also just export module.aurora so that you can get all the attributes.

Oh let me try that!

PR #514, opened 2023-06-26
feat: CDI-1607 Add KMS encryption key var to s3 bucket

jayengee:
### Summary
Add functionality to set private s3 bucket encryption to use a specified (new) KMS key instead of the default SSE-S3 keys,  allowing cross-account access to the bucket without needing to assume a role in the target AWS account

### Test Plan

### References

### DIFF
--- a/aws-s3-private-bucket/main.tf
+++ b/aws-s3-private-bucket/main.tf
@@ -192,7 +192,12 @@ resource "aws_kms_key" "bucket_kms_key" {
   description              = "This key is used to encrypt bucket objects for bucket ${var.bucket_name}"
   customer_master_key_spec = var.kms_key_type
   tags                     = {
-    
+    project   = var.project
+    env       = var.env
+    service   = var.service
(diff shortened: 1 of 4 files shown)


aws-s3-private-bucket/main.tf
```diff
@@ -185,3 +185,26 @@ resource "aws_s3_bucket_ownership_controls" "bucket" {
...
 }
+
+resource "aws_kms_key" "bucket_kms_key" {
+  count = var.kms_encryption != null ? 1 : 0
+
+  description              = "This key is used to encrypt bucket objects for bucket ${var.bucket_name}"
+  customer_master_key_spec = var.kms_key_type
+  tags                     = {
+    
+  }
```
alldoami, line 196: Missing tags?

jayengee: doh

alldoami (approved):

PR #516, opened 2023-07-17
fix: Do not set S3 kms encryption by default

jayengee:
### Summary
Previous change set s3 KMS encryption by default

### Test Plan

### References


jayengee:
@kuannie1 - yes. This is specifically KMS encryption (which I introduced in July as an option), rather the Server Side Encryption (SSE). KMS encryption is great in that it allows cross-account bucket access, but also creates overhead as anyone accessing that bucket would also require access to the KMS key

This PR doesn't remove encryption but rather just switches the default protocol back to SSE

jakeyheath (approved):

PR #532, opened 2023-10-31
feat: CDI-2183 Add databricks-cluster-log-permissions module

jayengee:
### Summary
Copy `databricks-cluster-log-permissions` from shared-infra to cztack

### Test Plan

### References

### DIFF
--- a/databricks-cluster-log-permissions/providers.tf
+++ b/databricks-cluster-log-permissions/providers.tf
@@ -1,0 +1,10 @@
+provider "aws" {
+  alias  = "logs_destination"
+  region = var.destination_account_region
+
+  assume_role {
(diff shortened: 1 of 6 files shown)


databricks-cluster-log-permissions/providers.tf
```diff
@@ -0,0 +1,10 @@
+provider "aws" {
```
jayengee, line 1: I *think* this will work

Original: https://github.com/chanzuckerberg/shared-infra/blob/95541fe5d66de9f180303236fed141d65d2ce33c/terraform/modules/databricks-cluster-log-permissions/providers.tf

fctczi (approved):

PR #550, opened 2024-01-18
chore: [ONCALL-693] Add account-level groups to default Databricks policies

aldengolab:
### Summary
This PR gives account-level policy groups in Databricks permissions to use the associated workspaces where the policies are enacted.

### Test Plan
Will apply using ie databricks to make sure the permissions are properly loaded, then will upgrade all remaining databricks policy modules.


jayengee (approved):
Have also checked that these groups exist in the account UI

PR #579, opened 2024-04-23
feat: Add a new single node policy to databricks-default-cluster-policies

jayengee:
### Summary
Add new single node job policy to databricks-default-cluster-policies

### Test Plan

### References


jjb007 (approved):
Did you want to enforce a non-empty string as a prefix so as to not confuse with the pre-existing Databricks policies?

PR #614, opened 2024-08-14
feat: Create databricks catalog ext loc modules

alldoami:
### Summary
This creates a bucket for databricks catalogs to use. It sets up the iam permissions that allows databricks access to read/write to the bucket. In order to associate a databricks catalog to a S3 bucket, it needs a storage credential which is connected to an external location (which is the S3 bucket). 

The `databricks-catalog` module creates a new catalog associated to the bucket with the desired permissions.

### Test Plan
Tested modules in [shared-infra](https://github.com/chanzuckerberg/shared-infra/pull/9973). Will be removing from shared-infra after merging and releasing.


databricks-catalog-external-location/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "Name of the role for instance profiles"
+  type        = string
+}
+
+variable "name_prefix" {
+  description = "Prefix to apply to resources"
+  type        = string
+}
+
+variable "group_names" {
```
jayengee, line 21: Let's change this to `all_privileges_grantees` or equivalent, given the grants for these groups

databricks-catalog-external-location/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+  description = "Name of the bucket"
+  type        = string
+}
+
+variable "databricks_external_location_iam_role" {
+  description = "Name of the role for instance profiles"
+  type        = string
+}
+
+variable "name_prefix" {
```
jayengee, line 16: this seems to only be used for the external location name - is the expectation that we just set this to the catalog name?

databricks-catalog/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+variable "isolation_mode" {
+  description = "Isolation mode of the catalog"
+  type        = string
+  default     = "ISOLATED"
+}
+
+variable "catalog_owner" {
+  description = "Owner of the catalog"
+  type        = string
+  default     = ""
```
jayengee, line 20: Probably shouldn't be allowed to be blank?

databricks-catalog/main.tf
```diff
@@ -0,0 +1,52 @@
+resource "databricks_catalog" "catalog" {
```
jayengee, line 1: Given these are single bucket catalogs (we're not supporting prefixes), why split these two modules out? In case we do want to add prefix support later?

alldoami: I thought it'd be more straightforward if it were separated but I decided to create a list of maps so that it can all be passed in.

databricks-catalog-external-location/versions.tf
```diff
@@ -0,0 +1,12 @@
+terraform {
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+    databricks = {
+      source = "databricks/databricks"
+      version = "1.49.1"
```
alldoami, line 8: required for the `isolation_mode` attribute for `databricks_external_location`

databricks-catalog-external-location/catalogs.tf
```diff
@@ -0,0 +1,54 @@
+resource "databricks_catalog" "catalog" {
+  for_each       = { for idx, catalog in var.catalogs : catalog.name => catalog }
+  name           = each.value.name
+  storage_root   = "s3://${module.catalog_bucket.name}"
```
jayengee, line 4: I don't think you can/should have multiple catalogs pointing to the exact same storage root. Maybe add a prefix for the catalog name:
```
"s3://${module.catalog_bucket.name}/{each.value.name}"
```

databricks-catalog-external-location/main.tf
```diff
@@ -0,0 +1,169 @@
...
+resource "databricks_storage_credential" "external" {
+  depends_on = [
+    resource.aws_iam_role.databricks_external_location_iam_role,
+    module.catalog_bucket
+  ]
+  name = aws_iam_role.databricks_external_location_iam_role.name
+  aws_iam_role {
+    role_arn = aws_iam_role.databricks_external_location_iam_role.arn
+  }
+  isolation_mode = "ISOLATION_MODE_ISOLATED"
```
jayengee, line 126: Might want to set this to a VAR in case we need to allow cross-workspace access to a catalog, for some reason

databricks-catalog-external-location/main.tf
```diff
@@ -0,0 +1,169 @@
...
+}
+
+resource "databricks_external_location" "external_locations" {
+  depends_on = [
+    resource.databricks_storage_credential.external
+  ]
+  name            = local.external_location_name
+  url             = "s3://${local.bucket_name}"
+  credential_name = databricks_storage_credential.external.id
+  isolation_mode  = "ISOLATION_MODE_ISOLATED"
```
jayengee, line 154: Same as above

databricks-catalog-external-location/main.tf
```diff
@@ -0,0 +1,169 @@
+data "aws_caller_identity" "current" {
+  provider = aws
+}
+
+locals {
+  path                   = "/databricks/"
+  name                   = "${var.tags.project}-${var.tags.env}"
```
jayengee, line 7: I think you'll want to allow folks to pass in a string as well, in case we need to set up more than one external location per workspace

That's already the case for Edu, for example

databricks-catalog-external-location/main.tf
alldoami: ya what's bad about this variable is that it's all or nothing... like if it's not isolated, then it's open to all workspaces. there's not feature in the provider yet to select not all workspaces. is that ok?

databricks-catalog-external-location/catalogs.tf
alldoami: Hmm I'm pretty sure it is ok... the root is the same but then the location is different since each catalog gets put into a different folder within the root. I'll dm you some examples.

databricks-catalog-external-location/main.tf
jayengee: Ah right, yep - this is the right call. We can add bindings later

databricks-catalog-external-location/catalogs.tf
jayengee: Hmm, I still think an additional prefix would be a good idea, if for no other reason (assuming no uuid collision) at least making it easier to navigate

jayengee (approved):

PR #659, opened 2024-10-17
fix: Move out power user group

alldoami:
This group was causing issues if you wanted to create multiple versions of this module in one env.

### DIFF
--- a/databricks-default-cluster-policies/variables.tf
+++ b/databricks-default-cluster-policies/variables.tf
@@ -38,4 +38,5 @@ variable "policy_map" {
 variable "power_user_group_name" {
   description = "Name of the power user group"
   type = string
+  default = "Power Users"
 }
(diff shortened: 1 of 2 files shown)


databricks-default-cluster-policies/variables.tf
```diff
@@ -34,3 +34,8 @@ variable "policy_map" {
   description = "Map of policy names to groups"
   type = list(map(list(string)))
 }
+
+variable "power_user_group_name" {
```
jjb007, line 38: Do you feel a default could be used to keep the naming somewhat standard over all workspaces as a default, or is it okay for workspaces to explicitly create their own name?

jjb007 (approved):

databricks-default-cluster-policies/variables.tf
alldoami: yup I can put a default value in.

PR #704, opened 2025-03-05
chore: Allow limited custom name segments for Databricks Catalog reso…

jjb007:
…urces

Allows for a little bit of customization of S3 bucket and assumed role for access.

### DIFF
--- a/databricks-catalog-external-location/variables.tf
+++ b/databricks-catalog-external-location/variables.tf
@@ -18,6 +18,22 @@ variable "external_location_name" {
   type        = string
 }
 
+variable "override_bucket_name_prefix" {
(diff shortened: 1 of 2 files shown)


databricks-catalog-external-location/variables.tf
```diff
@@ -18,6 +18,22 @@ variable "external_location_name" {
   type        = string
 }
 
+variable "override_bucket_name_prefix" {
```
jayengee, line 21: I think this would be cleaner if you set this to default null, then adjusted the above logic to:
```
var.override_bucket_name_prefix != null ? <a> : <b>
```

jayengee (approved):

PR #711, opened 2025-03-19
feat!: Allow separate backing buckets for Databricks Catalog and Volume

jjb007:
### Summary
databricks-s3-volume:
- Add the ability to configure the backing buckets each for the Databricks Catalog and Databricks Volume
- Changes input variables for consistency

databricks-s3-volume will no longer be backwards-compatible.

Changes in other modules were just from updating Terraform formatting and/or docs.

### Test Plan
Tested in a TF workspace

### DIFF
--- a/databricks-s3-volume/main.tf
+++ b/databricks-s3-volume/main.tf
@@ -2,25 +2,34 @@
 
 // https://docs.databricks.com/administration-guide/multiworkspace/iam-role.html#language-Your%C2%A0VPC,%C2%A0custom
 locals {
-  unity_aws_role_name = "${var.catalog_name}-unity"
+  unity_aws_role_name    = "${var.catalog_name}-unity"
+  iam_role_path          = "/databricks/"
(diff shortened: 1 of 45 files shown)


databricks-s3-volume/main.tf
```diff
@@ -2,75 +2,119 @@
...
-
-  # Allow overriding the storage location in case of an existing bucket
-  storage_location = var.override_storage_location != null ? var.override_storage_location : "s3://${local.bucket_name}/${local.schema_name}/${local.volume_name}"
+  create_storage_credentials = var.create_catalog || var.create_storage_credentials
+  volume_storage_location = coalesce(
+    var.volume_storage_location,
+    "s3://${local.volume_bucket_name}/${local.schema_name}/${local.volume_name}"
+  )
 
+  dbx_resource_storage_config = toset([
```
jayengee, line 28: Doesn't this mean that we'd always create a catalog bucket and a volume bucket?

jjb007: Ensured that catalog bucket can be created only when specified. Also moved a bunch of iterators to locals for clarity and converted some to maps.

jayengee (approved):

PR #728, opened 2025-05-15
fix: Assign all listed Databricks usergroups to associated Databricks cluster policy groups

jjb007:
### Summary
Previously, as a list of usergroups was associated per cluster policy, only the first usergroup in the usergroup list was being added to the cluster policy's group membership.

Instead, "flattening" out to a list of policy-to-group mappings, and creating `databricks_group_member`ship for each explicitly. Each `databricks_group_member` resource is now named after the pairing instead of just the policy name.

### Test Plan
Tested in local TF workspace.

### DIFF
--- a/databricks-default-cluster-policies/groups.tf
+++ b/databricks-default-cluster-policies/groups.tf
@@ -5,9 +5,11 @@ locals {
 
   policy_group_membership_list = toset([
     for prefix in local.ws_policy_name_prefixes
-    : for policy_suffix, groups_names in var.policy_map
-      : for group_name in groups_names
+    : [ for policy_suffix, groups_names in var.policy_map
+      : [ for group_name in groups_names
         : "${prefix}${policy_suffix}" => "${group_name}"
+      ]
+    ]
   ])
 }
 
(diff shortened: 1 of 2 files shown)


databricks-default-cluster-policies/groups.tf
```diff
@@ -3,44 +3,57 @@ locals {
...
 
-# Assign the existing groups to the newly created policy groups
+# Assign user groups to policy groups per policy-usergroup pair
 resource "databricks_group_member" "ws_policy_group_members" {
-  for_each = databricks_group.ws_policy_groups
-
-  group_id = each.value.id
-
-  # Assign all existing groups that correspond to this policy group
-  member_id = data.databricks_group.groups[local.policy_group_map[replace(each.key, var.policy_name_prefix, "")][0]].id
```
jjb007, line 45: Indexing 0 was retrieving only the first group member in the list per policy.

jayengee: Yikes :(

jayengee (approved):

PR #746, opened 2025-09-18
fix: renamed github action role for terraform test

naihsuanshao:
### Summary
This PR is for fixing the IAM role name for terraform test
Screenshot 2025-09-18 at 3 10 04 PM

### Test Plan

### References


naihsuanshao:
> Is it okay that 471112759938 account is non-prod?

I think it should be fine. We restrict this role to only be able to deploy in the core-platform-nonprod account

PR #914, opened 2026-08-17
fix!: tolerate the cilium startup taint in Deployment-backed EKS addons

alexlokshin-czi:
## Summary

Two related changes for fresh-cluster bootstrap and day-2 ownership of the default Karpenter resources.

### 1. Cilium startup-taint bootstrap tolerations

`karpenter_declare_cilium_startup_taint = true` deadlocks a fresh cluster: every Karpenter node carries `node.cilium.io/agent-not-ready:NoSchedule` from birth, but the cilium agent that removes it is installed via ArgoCD only after the cluster is registered — which happens after this module applies (first hit: prod-katamari, chanzuckerberg/argus-infra-stacks#2967). The flag now injects a matching toleration into everything the apply hard-waits on:

* **Stage-2 EKS addons** — `coredns` (`tolerations`), `aws-ebs-csi-driver` and `aws-mountpoint-s3-csi-driver` (`controller.tolerations`) via `configuration_values`.
* **external-secrets** — via the chart's `global.tolerations` (this module's config defaults `wait = true`).

DaemonSet-backed addons already ship `{operator: Exists}` tolerations; remaining stage-3 releases default `wait = false` and self-heal once cilium lands. Uncovered exceptions (`kubecost`, `jupyterhub`, both default-off) documented in the variable description. Requires Kubernetes 1.28+ (older s3-csi addon schemas reject `controller.tolerations`). All toleration payloads derive from one `cilium_addon_configuration_values` local next to the taint definition; chart-default tolerations were verified byte-exact against live `describe-addon-configuration` schemas, and external-secrets `global.tolerations` rendering via `helm template`.

### 2. Terraform creates the default NodePool/EC2NodeClass once; ArgoCD owns them afterwards

Per @jakeyheath's review comment. Terraform must create these at bootstrap (stage 2/3 addons need compute before the cluster is registered with ArgoCD), but afterwards ArgoCD manages updates — non-destructively, via Karpenter drift reconciliation, instead of this module's replace-the-pool-and-drain-everything model:

* The spec-hash-named pool (`random_id.node_pool_name` keepers + `force_new`) is replaced by a stable **`NodePool/default`** at a **new resource address** (`kubectl_manifest.karpenter_default_nodepool`). A new address is deliberate: adding `ignore_changes` to the existing resource while renaming it would suppress the rename itself, freezing the hash name forever and leaving ArgoCD to create a second pool instead of adopting.
* Both handed-off objects get the two mechanisms that are each individually insufficient: `lifecycle { ignore_changes = [yaml_body] }` freezes the Terraform config side, and `ignore_fields = ["metadata.labels", "metadata.annotations", "spec"]` blinds the provider's in-cluster drift fingerprint. Without the latter, ArgoCD's first spec change would trigger a re-apply of the stored `yaml_body` and a permanent revert fight (`ignore_fields` affects only drift detection, never the applied payload — verified against provider source at v1.19.0).
* `server_side_apply = true` + `force_conflicts = true`: creation is an idempotent upsert, so the recommended migration order (enable the ArgoCD app first, pool pre-exists, then bump) has no capacity gap; ArgoCD's SSA (gitops-engine forces conflicts) reclaims field ownership on its next sync. `create_before_destroy` and `force_new` are dropped — CBD plus a stable name plus upsert-create would let a future replacement delete the live object out from under its own new state entry.
* The ODCR EC2NodeClass stays fully Terraform-managed (its selector terms are TF vars operators actively change; stable name, updates in place).

`karpenter_nodepool_spec`, `karpenter_declare_cilium_startup_taint` (its NodePool half), `karpenter_ami_selector_terms`, and `docker_storage_size` become creation-time-only inputs for these two objects — post-creation updates happen via the ArgoCD values. The toleration half of the taint flag still applies on every apply. This also dissolves the original hazard of this PR: flipping the taint flag no longer touches the NodePool at all after creation.

## Rollout

All 56 module consumers pin the floating `aws-eks-cluster-v8` tag, so this lands as **v9** and nothing changes for anyone until a repo bumps its ref — the bump is the per-repo rollout gate. **The bump drains nothing**: the legacy hash-named pool is kept and neutered in place with `limits.cpu: 0` (its name frozen via `ignore_changes` on the `random_id` keepers; limits are excluded from Karpenter's drift hash, so existing nodes are untouched while new launches are blocked), and capacity migrates to the new `NodePool/default` through normal `expireAfter` rotation and consolidation under disruption budgets (full migration within ~15 days, typically sooner). Expected plan on bump: create `NodePool/default`, in-place update of the legacy pool (adds limits), in-place update of the nodeclass; **no destroys** — anything else is a red flag. Once a cluster's legacy pool reports zero NodeClaims, set `karpenter_legacy_nodepool = false` to delete the then-empty pool; fresh clusters can set it `false` at creation to never grow the legacy pool. Operator warnings live in the `addons` variable description: never enable the default pool on a cluster that already has a NodePool named `default` (the SSA-force upsert silently overwrites it — tfe-prod/tfe-staging carry such pools today with the module pool disabled); don't bump before the cluster is hub-registered with the ArgoCD app available (or the pool freezes with no updater); disable the ArgoCD app before turning off the enable flags (deletion is otherwise self-healed back).

## Verification

* `terraform fmt` clean; `terraform validate` produces no new diagnostics vs the branch base; zero remaining references to the deleted addresses.
* kubectl provider mechanics verified from source at v1.19.0; the fleet's oldest pin (1.14.0, hub clusters) documents all four attributes used.
* Design informed by a live inventory of all 54 reachable fleet clusters and a sweep of all 56 consumers: module-created pools carry no ArgoCD tracking metadata (clean adoption); per-cluster deviations to preserve in ArgoCD values recorded (prod-bhp on-demand-only + AMI pin, prod-mga AMI pin, 150Gi volumes on si-playground/single-cell-curation).

BREAKING CHANGE: the default NodePool/EC2NodeClass become creation-time-only in Terraform and are owned by ArgoCD after creation; upgrading creates the stable `default` NodePool and freezes provisioning on the legacy hash-named pool (limits.cpu=0, in-place, no node drained) — capacity migrates via normal expireAfter rotation, and the emptied legacy pool is removed by setting karpenter_legacy_nodepool = false.


jakeyheath:
re what we spoke about in person. Use ignore changes so that argocd can take over ownership of this resource and update the nodepool

alexlokshin-czi:
> re what we spoke about in person. Use ignore changes so that argocd can take over ownership of this resource and update the nodepool

I am going to keep to keep addon tolerations relaxed regardless so they won't be blocked when new nodes are created; this is not limited to a default nodepool created in terraform.

alexlokshin-czi:
Done in f5739e4 — went with ignore_changes plus the provider's `ignore_fields`, both are needed: `ignore_changes = [yaml_body]` only freezes the config side, while the kubectl provider re-applies the stored yaml_body whenever a field it set drifts in-cluster (the diff lands on the `yaml_incluster` fingerprint), so `ignore_fields = ["metadata.labels", "metadata.annotations", "spec"]` is what actually keeps Terraform from fighting ArgoCD.

Two deliberate choices worth calling out:
- New resource address with a stable name (`NodePool/default`) instead of ignore_changes on the existing one — otherwise the rename itself gets suppressed and existing clusters keep the frozen hash-named pool while ArgoCD creates a second one instead of adopting.
- SSA upsert with `force_conflicts = true`, so the safe migration order is ArgoCD-first: enable the app, `default` pre-exists, then bump the module ref — the drain from the old pool lands on capacity that's already there.

EC2NodeClass `default` gets the same treatment; the ODCR nodeclass stays fully TF-managed. Ships as v9 (BREAKING CHANGE footer), so nothing moves until each repo bumps off the floating v8 tag.

jakeyheath (approved):

hspitzley-czi (commented):
is `fix!` going to cause a major version bump? if so, it will require bumping all callers, is that what we want?

alexlokshin-czi:
> is `fix!` going to cause a major version bump? if so, it will require bumping all callers, is that what we want?

Yes, this is a breaking change, and we will need to update all callers over time.

hspitzley-czi (approved):

PR #915, opened 2026-08-19
feat(aws-eks-cluster): cluster_tags passthrough for per-cluster tags

alexlokshin-czi:
## Summary

Exposes the upstream `terraform-aws-modules/eks` module's `cluster_tags` input (map, default `{}` — additive tags on the EKS cluster resource only). Motivation (CCIE-7191): GuardDuty's org-level EKS addon management force-installs the runtime agent into clusters whose VPCs lack a `guardduty-data` endpoint, leaving 48-pod DaemonSets in permanent NXDOMAIN crashloop (~13k restarts/day on prod-central alone). The documented opt-out is a `GuardDutyManaged=false` tag on the cluster — which must live in Terraform or the next apply strips it. A follow-up in argus-infra-stacks sets the tag on the five affected clusters.

No behavior change for existing consumers (default `{}`).

## Test plan

- `terraform validate` passes; upstream input verified present in eks module 19.16.0.
- Consumer PR (argus-infra-stacks) will show a tags-only in-place update on `aws_eks_cluster`.

### DIFF
--- a/aws-eks-cluster/main.tf
+++ b/aws-eks-cluster/main.tf
@@ -154,6 +154,8 @@ module "cluster" {
 
   tags = merge(var.tags, local.karpenter_discovery, local.karpenter_discovery_per_cluster)
 
+  cluster_tags = var.cluster_tags
+
   cluster_name              = local.cluster_name
   cluster_version           = local.cluster_version
   vpc_id                    = var.vpc_id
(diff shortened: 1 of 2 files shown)


alexlokshin-czi:
Closing: resolved operationally instead. GuardDuty was excluded on the five affected clusters directly — `GuardDutyManaged=false` tag via `aws eks tag-resource` plus deletion of the orphaned `aws-guardduty-agent` addon objects (the crashlooping namespaces were already removed by hand). Caveat noted on CCIE-7191: the CLI tag lives outside Terraform, so a future `eks` workspace apply may strip it (the module manages cluster tags); if the crashloop ever returns after an eks apply, re-tagging — or reviving this passthrough — is the fix.

